PR #4, opened 2018-07-16
Fix KVS regex to correctly filter path-unsafe characters

mnmkng:
A supplemental change that is needed for fixing https://github.com/apifytech/apify-js/issues/123
PR: https://github.com/apifytech/apify-js/pull/125

It also fixes Jira:AP-477

_It's a breaking change_


mtrunkat:
Please add test to `test/regexs`

PR #5, opened 2018-07-26
Add helper function to serialize, compress and chunk data to be transferred over network 

mnmkng:
Implements https://github.com/apifytech/apify-js/issues/70 as a helper function that will initially be used in `apify-client/datasets`.

### DIFF
--- a/src/utilities.js
+++ b/src/utilities.js
@@ -451,21 +451,17 @@ const gzipP = Promise.promisify(gzip);
 exports.stringifyGzipChunkArray = function stringifyGzipChunkArray(array, maxChunkSizeBytes) { // using "function" to be able to self-reference
     if (!array.length) return [];
     const serialized = JSON.stringify(array);
-    return gzip(serialized).then((zipped) => {
+    return gzipP(serialized).then((zipped) => {
         const byteSize = Buffer.byteLength(zipped);
         // return fast if size is ok
         if (byteSize < maxChunkSizeBytes) return [zipped];
         // now we know that it isn't, first check if we can do anything about it
-        if (array.length === 1) throw new Error(`Unable to serialize: Chunk too large! Item on index: 0 is ${byteSize} bytes.`); // eslint-disable-line
+        if (array.length === 1) throw new Error(`Unable to serialize: Chunk too large! Item is ${byteSize} bytes. Limit is ${maxChunkSizeBytes} bytes.`); // eslint-disable-line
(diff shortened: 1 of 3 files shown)


src/utilities.js
```diff
@@ -437,3 +438,43 @@ exports.promisifyServerListen = (server) => {
...
+    const serialized = JSON.stringify(array);
+    return gzipP(serialized).then((zipped) => {
+        const byteSize = Buffer.byteLength(zipped);
+        // return fast if size is ok
+        if (byteSize < maxChunkSizeBytes) return [zipped];
+        // now we know that it isn't, first check if we can do anything about it
+        if (array.length === 1) throw new Error(`Unable to serialize: Chunk too large! Item is ${byteSize} bytes. Limit is ${maxChunkSizeBytes} bytes.`); // eslint-disable-line
+        if (array.length > 1 && Math.ceil(byteSize / array.length) > maxChunkSizeBytes) throw new Error(`Unable to serialize: Data too large! Array cannot be split into chunks < ${maxChunkSizeBytes} bytes.`); // eslint-disable-line
+        // theoretically possible to chunk now, let's assume that all items are roughly equal in size
+        const numOfChunks = Math.ceil(byteSize / maxChunkSizeBytes);
```
mtrunkat: I would chunk it to lets say half (constant) of maxChunkSizeBytes. Because of this way you very likely excess maxChunkSizeBytes again.

mtrunkat: Or even third. If max payload is 9MB then making requests for 3MB is not much overhead.

src/utilities.js
```diff
@@ -437,3 +438,43 @@ exports.promisifyServerListen = (server) => {
...
+        const maxItemsPerChunk = Math.floor(array.length / numOfChunks);
+        const chunked = _.chunk(array, maxItemsPerChunk);
+        // attempt to serialize the chunks
+        const zipPromises = chunked.map(chunk => stringifyGzipChunkArray(chunk, maxChunkSizeBytes));
```
mtrunkat: You can replace these 2 lines by: 
(might be better to do it one by one since we don't know like big is the array)

```
return Promise.mapSeries(chunked, chunk => stringifyGzipChunkArray(chunk, maxChunkSizeBytes))
    .catch(...
```

mtrunkat: I think that we should limit the depth of recursion to something meaningful like 100. If you have an array o N items and one of them is larger then 9MB then it gets log_2(N) deep in the stack before it throws an error. For very large array it could exceed max stack size.

mnmkng:
Closing this in favor of https://github.com/apifytech/apify-js/pull/138

PR #7, opened 2018-08-03
Move list of User-Agent headers from apify-system to apify-shared

mnmkng:
A list of `User-Agent` headers is necessary for the `rotateUserAgents` Crawler option.
Since we're migrating Crawler to Actor (e.g. `crawler-cheerio`),
it makes sense to keep relevant constants in one place only.

I downloaded a fresh list off the page in comment and will prepare a PR in `apify-system`
that would use this list from `apify-shared`.


mnmkng:
Closing this. Crawler cannot load constants from npm package. Adding the `User-Agent` list to `apify-js`.

PR #9, opened 2018-08-31
Refactoring local env variables

mtrunkat:


src/consts.js
```diff
@@ -282,6 +253,30 @@ export const ENV_VARS = {
     CONTAINER_URL: 'APIFY_CONTAINER_URL',
 };
 
+/**
+ * Local emulation sub directories for local stores
+ */
+export const LOCAL_STORAGE_SUBDIRS = {
+    datasets: 'datasets',
+    keyValueStores: 'key-value-stores',
+    requestQueues: 'request-queues',
```
mnmkng: This is same as current, but shouldn't we use either all underscores or all dashes?

`apify_storage` vs `key-value-stores`

Makes it difficult to write paths without checking syntax first.

It looks good otherwise! Are you gonna do the SDK PR that sets the variables too?

mtrunkat:
Good point!

pá 31. 8. 2018 v 19:14 odesílatel Ondra Urban <notifications@github.com>
napsal:

drobnikj:
need some changes in CLU, I created issue for that https://github.com/apifytech/apify-cli/issues/52 .

PR #15, opened 2019-01-17
Fixed log.skipLevelInfo implementation

jancurn:


mtrunkat:
@mnmkng, where is this used?

jancurn:
It's not used anywhere, so it should be safe.

jancurn:
Guys so can we merge this and publish new version?

PR #21, opened 2019-03-20
Input schema: Added custom proxy url validation

gippy:
Also added option passed from application which controls whether user has any proxy groups available for automatic mode (now it does not validate this properly, because every user has access to residential and serp, but they are excluded from auto mode)


src/intl.js
```diff
@@ -8,16 +8,20 @@ const intlStrings = {
     'inputSchema.validation.objectValuesInvalid': 'Keys [{invalidKeys}] in input.{fieldKey} should have string value which matches pattern "{pattern}"',
     'inputSchema.validation.additionalProperty': 'Property {fieldKey} is not allowed.',
     'inputSchema.validation.proxyGroupsNotAvailable': 'You currently do not have access to proxy groups: {groups}',
+    'inputSchema.validation.customProxyInvalid': 'Proxy URL {invalidUrl} is not valid.',
```
mtrunkat: We could udpate message `Proxy URL {invalidUrl} doesn't have valid format http://[URL]:[PORT]

PR #25, opened 2019-05-08
Add WebhookPayloadTemplate class to handle curly braces webhook syntax

mnmkng:
Title says all. Putting it here since it's gonna be useful in SDK too when creating webhooks programmatically.

Currently does not support dot notation `{{var.prop}}`, but I'll add it later.

### DIFF
--- a/src/webhook_payload_template.js
+++ b/src/webhook_payload_template.js
@@ -36,6 +36,7 @@ const { JsonToken, jsonStringifyExtended } = require('./utilities.client');
  *    }
  * }
  * ```
+ * @hideconstructor
  */
 export default class WebhookPayloadTemplate {
     constructor(payloadTemplate, allowedVariables = null, context = {}) {
@@ -54,8 +55,8 @@ export default class WebhookPayloadTemplate {
      * Parse also validates the template structure, so it can be used
      * to check validity of the template JSON and usage of allowedVariables.
      * @param {string} payloadTemplate
-     * @param {Set<string>|null} [allowedVariables]
-     * @param {Object} [context]
+     * @param {Set<string>|null} [allowedVariables=null]
+     * @param {Object} [context={}]
      * @return {object}
      */
     static parse(payloadTemplate, allowedVariables, context) {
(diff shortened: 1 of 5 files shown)


src/webhook_payload_template.js
```diff
@@ -0,0 +1,164 @@
+const { JsonToken, jsonStringifyExtended } = require('./utilities.client');
```
jancurn: Wasn't this supposed to be called `JsonVariable` ?

src/webhook_payload_template.js
```diff
@@ -0,0 +1,164 @@
...
+     *     "bar": ["bar"],
+     *     "res": {{resource}},
+     * }
+     * ```
+     *
+     * @param {string} variable
+     * @return {JsonToken}
+     */
+    static getVariable(variable) {
+        return new JsonToken(variable);
```
jancurn: Is it really necessary to define such a simple function for this?

src/webhook_payload_template.js
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * This is useful to create dynamic webhook payloads where a template is saved on webhook
+ * creation and then variables are dynamically added on webhook dispatch.
+ *
+ * **Example:**
+ * ```js
+ * const payloadTemplate = `{
+ *    "id": "some-id",
+ *    "createdAt": "2019-05-08T15:22:21.095Z",
+ *    "dataToSend": {{data}}
```
jancurn: From the comments it's not clear how to handle hierarchical objects, is it even possible? E.g. to use `{{data.status}}`

src/webhook_payload_template.js
mnmkng: Don't know, that was @mtrunkat's PR. I'm just using it.

src/webhook_payload_template.js
mnmkng: It's an encapsulation thing. You don't want to do

```
const { jsonToken, jsonStringifyExtended } = require('apify-shared/utilities.client');
const WebhookPayloadTemplate = require('apify-shared/webhook_payload_template');

const var = new jsonToken('var');
```

when you can just think about having the one class and not even know about the `jsonToken` which is something else entirely.

src/webhook_payload_template.js
mnmkng: It's not possible and it's in the OP of the PR.

src/webhook_payload_template.js
mtrunkat: I renamed that to `JsonVariable`

PR #31, opened 2019-12-23
Updating packages and linting

mtrunkat:


jancurn:
Shouldn't that be to `develop` rather than `master`?

PR #34, opened 2020-01-20
Added "noopener noreferrer" to rel in markedSetNofollowLinks

novotnyj:


src/utilities.client.js
```diff
@@ -154,7 +154,9 @@ export const markedSetNofollowLinks = (href, title, text) => {
         // Probably invalid url, go on
     }
     const isApifyLink = (urlParsed && /\.apify\.com$/i.test(urlParsed.hostname));
```
jancurn: Shouldn't this be `/(\.|^)apify\.com$/`, to cover also links like https://apify.com/xxx

PR #35, opened 2020-01-27
Add RUN_MEMORY_SIZE_TO_CPU_CORES_COEFF constant

fnesveda:
Add a constant containing the coefficient for converting the run memory to available CPU cores of the run.

Necessary for [https://bitbucket.org/apifyteam/apify-system/pull-requests/1881](https://bitbucket.org/apifyteam/apify-system/pull-requests/1881).

### DIFF
--- a/src/consts.js
+++ b/src/consts.js
@@ -212,7 +212,7 @@ export const ACTOR_LIMITS = {
     RUN_DISK_TO_MEMORY_SIZE_COEFF: 2,
 
     // For each build or run container, set CPU cores based on memory size
-    RUN_MEMORY_SIZE_TO_CPU_CORES_COEFF: 4096,
+    RUN_MEMORY_MBYTES_PER_CPU_CORE: 4096,
 
     // The default limit of memory for all running Actor jobs for free accounts.
     FREE_ACCOUNT_MAX_MEMORY_MBYTES: 8192,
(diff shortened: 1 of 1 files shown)


src/consts.js
```diff
@@ -211,6 +211,9 @@ export const ACTOR_LIMITS = {
     // For each build or run container, set disk quota based on memory size
     RUN_DISK_TO_MEMORY_SIZE_COEFF: 2,
 
+    // For each build or run container, set CPU cores based on memory size
+    RUN_MEMORY_SIZE_TO_CPU_CORES_COEFF: 4096,
```
jancurn: Just one note - we always thrive to make constants contain units, otherwise, Mars probes can crash (https://en.wikipedia.org/wiki/Mars_Climate_Orbiter#Cause_of_failure).

How about calling this `RUN_MEMORY_MBYTES_PER_CPU_CORE` ?

fnesveda: That makes sense, done in 645a128.

src/consts.js
```diff
@@ -212,7 +212,7 @@ export const ACTOR_LIMITS = {
     RUN_DISK_TO_MEMORY_SIZE_COEFF: 2,
 
     // For each build or run container, set CPU cores based on memory size
-    RUN_MEMORY_SIZE_TO_CPU_CORES_COEFF: 4096,
```
mtrunkat: Are you sure that this is not used anywhere? (web, docs, sdk, ... )?

jancurn: That would mean the change was faster than light :) See files changed

fnesveda: I'm sure, I added this const in the previous commit in this PR, and then just renamed it to another name per Honza's suggestion (and I've renamed it to the new name in the related PR on Bitbucket as well, which was the only place where it was used so far).

PR #36, opened 2020-02-04
Validate Apify proxy country

fnesveda:
Adds validation for the selected country for Apify proxy.

### DIFF
--- a/src/intl.js
+++ b/src/intl.js
@@ -22,6 +22,8 @@ const intlStrings = {
         'You currently do not have access to proxy groups: {groups}',
     'inputSchema.validation.customProxyInvalid':
         'Proxy URL "{invalidUrl}" has invalid format, it must be http://[username[:password]]@hostname:port.',
+    'inputSchema.validation.apifyProxyCountryInvalid':
+        'Country code "{invalidCountry}" is invalid.',
     'inputSchema.validation.noAvailableAutoProxy':
         'Currently you do not have access to any proxy group usable in automatic mode.',
     'inputSchema.validation.noMatchingDefinition':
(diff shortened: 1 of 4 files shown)


src/utilities.client.js
```diff
@@ -399,6 +400,11 @@ function validateProxyField(fieldKey, value, isRequired = false, options = null)
     // If Apify proxy is not used skip additional checks
     if (!useApifyProxy) return fieldErrors;
 
+    // If Apify proxy is used, check if there is a selected country and if so, check that it's valid (empty or a valid country code)
+    if (apifyProxyCountry && apifyProxyCountry !== '' && !countries[apifyProxyCountry]) {
```
jancurn: The `apifyProxyCountry !== ''` condition is redundant, the first one will cover this.

fnesveda: Fixed in fdb4a6c.

fnesveda:
@gippy That's a good point, added the check in 8c54d99.

src/intl.js
```diff
@@ -22,6 +22,10 @@ const intlStrings = {
         'You currently do not have access to proxy groups: {groups}',
     'inputSchema.validation.customProxyInvalid':
         'Proxy URL "{invalidUrl}" has invalid format, it must be http://[username[:password]]@hostname:port.',
+    'inputSchema.validation.apifyProxyCountryInvalid':
+        'Country code "{invalidCountry}" is invalid.',
```
mtrunkat: Let's metion in error that `ISO 3166-1 alpha-2 codes` are supported  https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2

PR #37, opened 2020-02-04
Improvements in error logging

jancurn:
Removed log.methodCall and log.methodException functions, improved SOFT_FAIL/ERROR decision logic in log.exception() to be more aligned with error codes defined in app.

This must only be merged after https://bitbucket.org/apifyteam/apify-system/pull-requests/1895 !!!

### DIFF
--- a/src/log.js
+++ b/src/log.js
@@ -162,25 +162,7 @@ const logSoftFail = function (message, data) { logInternal(message, data, LOG_LE
 
 const logException = function (exception, message, data) {
     if (!data) data = {};
-
-    // If it's Meteor.Error then don't print stack trace and use "SOFT_FAIL" level
-    // which is used for less serious errors.
-    if (exception && exception.errorType === 'Meteor.Error') {
-        const conciseException = {
-            code: exception.error,
-            reason: exception.reason || exception.message || exception.stack,
-            details: exception.details,
-        };
-
-        // Determine log level
-        const httpCode = (exception.details ? exception.details.httpCode : null) || 500;
-        const forceSoftFail = exception.details ? exception.details.forceSoftFail : false;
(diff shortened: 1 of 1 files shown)


src/log.js
```diff
@@ -163,15 +163,21 @@ const logSoftFail = function (message, data) { logInternal(message, data, LOG_LE
...
             details: exception.details,
         };
-        logInternal(message, data, LOG_LEVELS.SOFT_FAIL, conciseException);
+
+        // Determine log level
+        const httpCode = (exception.details ? exception.details.httpCode : null) || 500;
+        const forceSoftFail = exception.details ? exception.details.forceSoftFail : false;
+        const level = httpCode >= 500 && !forceSoftFail ? LOG_LEVELS.ERROR : LOG_LEVELS.SOFT_FAIL;
+
+        logInternal(message, data, level, conciseException);
```
mtrunkat: This could cause problems for users. I think that if you call `log.exception()` then it should be logged as an error. For example, in actor, you are calling our API, you can get rate limit error 429 or any other error from 4xx range, the API client throws with the error that contains details and HTTP code, handlePageFunction fails with that error and this would log it as SOFT_FAIL although it's an error. In this use case, it's an important error and mainly the soft fail is something internal that should not be used in Apify SDK and client as it's nonstandard.

I think that this logic should be handled as soft fall in-app and not in the logger as recognition of 4xx error as softfail is not generic thing and this logger is used in public projects.

mtrunkat: Maybe we could just extend the logger in app with this logic that's just for app, the new logger will support it.

mtrunkat: @mnmkng could you check this change? It affects SDK logging

jancurn: @mtrunkat @mnmkng No need, let's get rid of this special case from apify-shared completely, it really shouldn't be here.

PR #41, opened 2020-02-25
Fix empty objects in log

mnmkng:


package.json
```diff
@@ -64,6 +64,7 @@
     "jsdoc-export-default-interop": "^0.3.1",
     "jsdom": "^15.1.1",
     "mocha": "^7.0.1",
-    "sinon": "^8.0.1"
+    "sinon": "^8.0.1",
+    "strip-ansi": "^6.0.0"
```
mnmkng: It's included in `chalk`. This is just so `ESLint` does not complain.

PR #52, opened 2020-04-23
Adding health checker

mtrunkat:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,32 +1,3 @@
 {
-  "extends": "airbnb-base",
-  "plugins": [
-    "import",
-    "promise"
-  ],
-  "rules": {
-    "indent": ["error", 4, { "SwitchCase": 1 }],
-    "no-underscore-dangle": [2, {
-      "allow": ["_id", "_overrideMethodGroups"]
-    }],
-    "no-use-before-define": 0,
-    "no-param-reassign": 0,
-    "consistent-return": 0,
-    "no-buffer-constructor": 1,
-    "object-curly-newline": 0,
-    "function-paren-newline": 0,
-    "prefer-destructuring": 0,
-    "array-callback-return": 0,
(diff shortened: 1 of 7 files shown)


jancurn:
Looks good! I'm thinking, wouldn't it be better to have the setInterval functionality integrated in the class too? Otherwise each user of this class will have to implement it themselves. Basically you could make the class inherit EventEmitter and just send events "healthy" or "unhealthy", to which the user could react. BTW, do you want to consider MongoDB down after a single error? There might be just a temporary network issue, usually healthchecks require the condition to repeat a few times...

mtrunkat:
> Looks good! I'm thinking, wouldn't it be better to have the setInterval functionality integrated in the class too? Otherwise each user of this class will have to implement it themselves. Basically you could make the class inherit EventEmitter and just send events "healthy" or "unhealthy", to which the user could react. BTW, do you want to consider MongoDB down after a single error? There might be just a temporary network issue, usually healthchecks require the condition to repeat a few times...

I would not do this as a health check is in all our implementations an API endpoint so it's easier to just execute checks on request instead of listening for events and storing the state somewhere. So after how many fails we consider service to be down is managed higher in infrastructure - either by Kubernetes health check configuration or ELB health checks, ...

jancurn:
@mtrunkat Oh, I thought that his healtcheck will just restart the process - that's usually enough to regain connection to Redis or Mongodb. ELB/ASG will remove the instance, which will take many minutes to restart...

mtrunkat:
> @mtrunkat Oh, I thought that his health check will just restart the process - that's usually enough to regain connection to Redis or MongoDB. ELB/ASG will remove the instance, which will take many minutes to restart...

Good point. In containerized applications, it's handled correctly by Kubernetes' health-check configuration. But in an App and API that is not containerized yet I will do this call to exit manually for now.

PR #54, opened 2020-05-05
Fix unwanted whitespace after title text

m-murasovs:
The title text was being generated with a lot of unwanted whitespace at the end. 
@vbartonicek says it's not urgent, so fixed for next package release


mtrunkat:
Released as `v0.2.14`

PR #55, opened 2020-05-07
Add PLATFORM_LIMITS

m-murasovs:


src/consts.js
```diff
@@ -233,6 +233,29 @@ export const ACTOR_LIMITS = {
     LOG_MAX_CHARS: 5000000,
 };
 
+/**
+ * Contains various limits of the Apify platform.
+ */
+export const PLATFORM_LIMITS = {
```
jancurn, line 239: These are just defaults, so this should be called DEFAULT_PLATFORM_LIMITS, to make it clear

jancurn: Also, I assume we'll then use these constants in the app, right?

m-murasovs: Cool, thanks! 
Also, yep, @mtrunkat asked to add this

PR #58, opened 2020-05-16
Fixed image proxy client to do not use jQuery and jsdom

drobnikj:


src/image_proxy_client.js
```diff
@@ -62,18 +62,23 @@ export default class ImageProxyClient {
...
-        const self = this;
-        $html.find('img').each(function () {
-            const imageUrl = $(this).attr('src');
-            $(this).attr('src', self.generateUrl(imageUrl));
+    updateImagesInHtml(html) {
+        const allImgElements = html.match(/<\s*img[^>]*\/?>/g);
+        if (!allImgElements) return html;
+
+        allImgElements.forEach((img) => {
+            const srcMatch = img.match(/src="([^">]+)"/);
```
jancurn, line 72: What if the html uses single-quotes instead of double-quotes?

PR #62, opened 2020-06-03
Exported traverseObject() function and extended the transformation fu…

jancurn:
…nction to enable also transformation of values, bumped package.json version

### DIFF
--- a/src/utilities.client.js
+++ b/src/utilities.client.js
@@ -258,14 +258,18 @@ export const unescapePropertyName = function (name) {
 };
 
 /**
- * Traverses an object, creates a deep clone if requested and transforms object keys using a provided function.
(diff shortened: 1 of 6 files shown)


src/utilities.client.js
```diff
@@ -342,13 +346,13 @@ exports.unescapeFromBson = function (obj, clone) {
...
     try {
-        traverseObject(obj, false, (key) => {
+        traverseObject(obj, false, (key, value) => {
             const escapedKey = escapePropertyName(key);
             if (key !== escapedKey) {
                 isBad = true;
                 throw new Error();
             }
-            return key;
+            return [key, value];
```
mnmkng, line 355: Since the return type of this function switches from `string` to `array`, shouldn't we make this a breaking change and bump major instead? I also need to bump major in #59 so we could publish 4.0 with both changes.

PR #71, opened 2020-08-25
Markdown renderer for tabbed examples

mato93:
Extending `marked` with custom renderer that can handle code block specifying tabs with code in different languages. This will be later consumed by React component and used for rendering.


src/marked.js
```diff
@@ -0,0 +1,66 @@
...
+
+export const apifyMarked = (() => {
+    const renderer = new marked.Renderer();
+    renderer.heading = customHeadingRenderer;
+    renderer.code = (code) => {
+        if (!code.trim().startsWith('```marked-tabs')) {
+            return false;
+        }
+
+        const matchesIterator = matchAll(code, /<marked-tab header="(?<header>.*?)" lang="(?<lang>.*?)">(?<code>.*?)<\/marked-tab>/sg);
```
mtrunkat, line 15: Food for thought: I am just thinking that usage of comments could be more readable when viewing and editing source markdown files at Github https://monosnap.com/file/rLXwI5FnrNVfqlPMtSoCL2K3cmmtdV but I don't have a strong opinion on this :)

src/marked.js
```diff
@@ -0,0 +1,66 @@
...
+                            }
+                        }
+                        return line;
+                    })
+                    .join('\n');
+
+                return `"${header}": { "language": "${lang}", "code: "${tabCode}" }`;
+            })
+            .join(',\n');
+        return `<apify-code-tabs input="{\n${inputArg.replace(/"/g, '\\"')}\n}" />`;
```
mtrunkat, line 61: it could be useful to move `apify-code-tabs` to some constant as Vrata will need it in the website implementation

src/marked.js
mato93: I've tried it out and I don't see a clear advantage of comments. MD editors (stackedit online and bear offline) seem not to highlight comments. Also, comments has disadvantage that they don't have closing `>` tag (or some other closing tag) which is worse for matching. So I'd stick with `<>`.

mato93:
@mtrunkat PTAL. I've addressed comments and also consolidated tests to `marked.js` file.

mtrunkat (approved):

PR #81, opened 2020-10-27
Convert relative URLs to absolute in git READMEs

mhamas:


src/git.js
```diff
@@ -0,0 +1,51 @@
+import GitUrlParse from 'git-url-parse';
```
mtrunkat, line 1: This should be lowercased as it's a function now constructor

src/git.js
```diff
@@ -0,0 +1,51 @@
...
+    const repoNameParts = parsedRepoUrl.full_name.split('/');
+    const repoFullName = `${repoNameParts[0]}/${repoNameParts[1]}`;
+
+    // When user wants to specify a different branch during actor addition, they need
+    // to add #branchName to the end of git URL. Hence it ends up under 'hash' key here.
+    // Otherwise, we try to use the branch name from the build object. It should exist for all builds since roughly mid October 2020.
+    // Prior to that, all default branches were 'master', hence we default to that as a backup.
+    const branchName = parsedRepoUrl.hash || gitBranchName || 'master';
+
+    // Images in markdown have syntax ![alt text](image url)
```
mtrunkat, line 28: There could be also includes as HTML (for example when embedded in HTML table) - ``

mtrunkat (changes requested):

mhamas:
@mtrunkat I've addressed the comments and verified locally that this works fine.

gippy (approved):

mtrunkat (approved):

PR #87, opened 2020-11-16
ImageProxyClient: Use create-hmac package instead of crypto

fnesveda:
We're using the ImageProxyClient also in the browser, and because of it's `crypto` requirement we have to polyfill the whole `crypto` module, which is huge and adds a lot of size to the Javascript bundle.

Since we need only the `createHmac` method from `crypto` in ImageProxyClient, I've replaced the usage of `crypto` in ImageProxyClient with the `create-hmac` module, which implements just that one function equivalently, but is much smaller.

I've tested this locally and it works well, I'm getting the same proxied URLs as before.


jancurn:
Nice. But IMHO it would be preferable not to use `ImageProxyClient` on client and generate the HTML on server, to avoid bundling these modules to clients altogether. The speed of page loads directly affects ranking on Google.

jancurn (approved):

drobnikj (approved):

mtrunkat (approved):

mtrunkat:
@fnesveda what happened to this? :)

PR #89, opened 2020-11-24
Cleanup of the repository

mhamas:
This PR
- incorporates changes from `npm run lint:fix`
- adds some eslint disable comments into the code to fix warnings
- merges changelog of versions `0.5.3` and `0.5.4` since the former hasn't yet been deployed so it can be deployed together.


mhamas:
Has been already resolved

PR #93, opened 2020-12-16
Improve replacement of relative paths in git READMEs.

mhamas:
All paths not starting with "http://", "https://" or "www." are considered relative.
Such relative paths with or without leading ./ are replaced correctly.


fnesveda:
Closing as duplicate of #98

PR #102, opened 2021-02-10
Rework custom heading renderer in apify-marked, update marked to fix vulnerability

fnesveda:
We're not using the current way to pass custom IDs to headings anywhere right now, because the markup in the required format renders badly if not using `apifyMarked`.

This PR reworks the way custom IDs are passed to this:
```md
 ### [](#custom-id) Heading text
````
becomes
````html
 <h3 id="custom-id"><a href="#custom-id"></a>Heading text</h3>
````

The link is rendered empty, so it takes no space, but in Docs we're using it to render the "Copy to clipboard" link and icon, and we had to add it everywhere manually up until now.

The `apifyMarked` renderer is not used anywhere but in Docs, so this should not break anything, but I'm releasing this as a breaking change just in case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+0.6.0 / 2021/02/11
+==================
+- Reworked custom heading renderer in `apifyMarked`
+- Updated `marked` to 2.0.0 because of https://github.com/advisories/GHSA-4r62-v4vq-hr96
+
 0.5.9 / 2021/02/04
 ==================
 - Added `PROXY_STATUS_URL` environment variable to the constants file
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "apify-shared",
-  "version": "0.5.9",
+  "version": "0.6.0",
   "description": "Tools and constants shared across Apify projects.",
   "main": "build/index.js",
   "keywords": [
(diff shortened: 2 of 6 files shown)


m-murasovs (changes requested):
Super impressive :) 

I ran the renderer in Repl and found:

On input without an ID, it works perfect:
Screenshot 2021-02-11 at 09 25 15

Same when the input has an ID that's properly formed
Screenshot 2021-02-11 at 09 26 12

But when there's some irregularity, it does this. 
Screenshot 2021-02-11 at 09 27 09

Or, if you forget to use the # symbol, you get two links
Screenshot 2021-02-11 at 09 29 06

I guess without the # symbol, it gets considered part of the heading text?
Could there be some handling for this, like adding dashes in the custom ID and a # if there isn't one to start with?

fnesveda:
The third one, with the space, is sorta by design / just how Markdown works - `[](#custom vibes)` is not a legal markdown link, so it does not get rendered as such, and then the regular expression does not parse it properly. We could add another regular expression dealing with cases like this, but then it breaks the invariant that the rendered markdown should look the same when rendered through `apifyMarked` as when rendered without it, e.g. through GitHub (because there the broken link would not be parsed and would be rendered just as `[](#custom vibes)`.

The fourth one, without the fragment, I'm not so sure about. The whole point of this is that the markdown part before the heading text should be a legal link pointing to the heading, and it is pointing to the heading only with the `#` symbol at the start, so only then it should be considered a custom ID.

I'd keep both of these behaviors as is, I think it's conceptually cleaner.

m-murasovs:
@fnesveda alright then :)

m-murasovs (approved):

mtrunkat (approved):

PR #107, opened 2021-02-26
Update Docker images

mnmkng:
Deprecated old docker images and replaced them with new ones. Not sure if this will work out of the box, because previously, `puppeteer` was always included in `dependencies` of the SDK, but now, users need to add it to their `package.json`. Are there different `package.json` files for different images? We need to add `puppeteer` to some and `playwright` to other.


drobnikj (changes requested):
The problem can be with a single file. We are installing packages(npm install package1 package2) matches from all `require('')` in code. If the user didn't require playwrite or any other requested package for the docker image the run will fail.

drobnikj:
Wait, maybe I'm not right as we already install these packages in the base image as they are listed there as dep e.g. https://github.com/apify/apify-actor-docker/blob/master/node-playwright-chrome/Dockerfile.
I took it back, it can work this way.

drobnikj (approved):

src/consts.js
```diff
@@ -68,54 +68,109 @@ export const META_ORIGINS = {
...
+        displayName: 'Node.js 14 + Playwright + All Browsers on Ubuntu',
         copyChown: 'myuser:myuser',
     },
 
     // Deprecated:
-    // TODO: Keep the for some time and then migrate acts to recommended images.
+    // These are here because we made breaking changes in the client that could break existing single file actors.
+    // We will get a rid of this along with the whole single file logic.
     {
-        name: 'apify/actor-node-puppeteer',
```
mtrunkat, line 107: Shouldn't we keep these just renamed to deprecated? Otherwise, the building will break if somebody use it.

mtrunkat (changes requested):

src/consts.js
mnmkng: `apify/actor-node-puppeteer` images were deprecated 3 years ago. I think it's safe to remove them.

jancurn: How about a quick query to see if somebody is still building with them?

mnmkng: I did the query and there were only some free users using it with non-production actors. It seems that they stumbled upon the image in the list and were trying it out. => Better remove it.

PR #115, opened 2021-03-24
Add custom link and image rendering functions

m-murasovs:
Needed for: https://app.zenhub.com/workspaces/platform-team-5f6454160d9f82000fa6733f/issues/apify/apify-web/861

### DIFF
--- a/src/markdown_renderers.js
+++ b/src/markdown_renderers.js
@@ -1,5 +1,7 @@
 import gitUrlParse from 'git-url-parse';
 
+const relativeUrlRegex = new RegExp('^(?!www.|(?:http|ftp)s?://|[A-Za-z]:\\|//).*', 'i');
+
 export const formatHeadingId = (headingId) => {
     // Replace non-word characters with dashes
     headingId = headingId.toLowerCase().trim().replace(/[^\w]+/g, '-');
@@ -119,8 +121,7 @@ export const customLinkRenderer = (href, text, repoUrl, branchName) => {
         href = href.toLowerCase();
     }
     // Only target relative URLs, which are used to refer to the git repo, and not anchors or absolute URLs
-    const relativeUrlRegex = new RegExp('^(?:[a-z]+:)?//', 'i');
-    if (!relativeUrlRegex.test(href)) {
+    if (relativeUrlRegex.test(href)) {
         const urlPrefix = generateGitRepoUrlPrefix(repoUrl, branchName, href);
         href = `${urlPrefix}/${href}`;
     }
@@ -139,7 +140,8 @@ export const customLinkRenderer = (href, text, repoUrl, branchName) => {
  * @return {string}
 */
 export const customImageRenderer = (href, text, repoUrl, gitBranchName) => {
-    if (!href.startsWith('http') && !href.startsWith('ftp') && !href.startsWith('#')) {
+    // Only target relative URLs, which are used to refer to the git repo, and not anchors or absolute URLs
+    if (relativeUrlRegex.test(href)) {
         const urlPrefix = generateRawGitRepoUrlPrefix(repoUrl, gitBranchName);
         href = `${urlPrefix}/${href}`;
     }
--- a/src/markdown_renderers.js
+++ b/src/markdown_renderers.js
@@ -1,3 +1,6 @@
+function interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; } // eslint-disable-line
+const gitUrlParse = interopRequireDefault(require('git-url-parse'));
+
 export const formatHeadingId = (headingId) => {
     // Replace non-word characters with dashes
     headingId = headingId.toLowerCase().trim().replace(/[^\w]+/g, '-');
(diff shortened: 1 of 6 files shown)


src/markdown_renderers.js
```diff
@@ -49,3 +52,69 @@ export const customHeadingRenderer = (text, level, raw) => {
...
+    if (href.startsWith('#')) {
+        href = href.toLowerCase();
+    }
+
+    // Replace relative URLs in README with absolute ones pointing to the actor's repo
+    if (!href.startsWith('http') && !href.startsWith('ftp') && !href.startsWith('#')) {
+        href = `${urlPrefix}/${href}`;
+    }
+
+    return `<a href=${href}>${text}</a>`;
```
drobnikj, line 98: You should to add rel="nofollow noreferrer" att into links go out of apify.

drobnikj (changes requested):

src/markdown_renderers.js
```diff
@@ -49,3 +52,69 @@ export const customHeadingRenderer = (text, level, raw) => {
...
+ * Mainly for use in actor READMES
+ * Parses the actor's repo URL to extract the repo name and owner name.
+ * @param {string} href
+ * @param {string} text
+ * @param {string} repoUrl
+ * @param {string} branchName
+ * @return {string}
+*/
+export const customLinkRenderer = (href, text, repoUrl, branchName) => {
+    const urlPrefix = generateGitRepoUrlPrefix(repoUrl, branchName);
```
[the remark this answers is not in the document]
m-murasovs: I wanted to do this initially but the github URL uses either `/tree` if you're linking a folder or `/blob` if you're linking a file, which would need some additional parsing. Do you think it's worth it?

src/markdown_renderers.js
```diff
@@ -1,3 +1,6 @@
+function interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; } // eslint-disable-line
```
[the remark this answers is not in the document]
m-murasovs: Yeah, I wasn't sure about this one, I was looking at the built `git` file for inspiration and it had this. I'll revise

src/markdown_renderers.js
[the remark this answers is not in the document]
m-murasovs: OK, I did it :D

src/markdown_renderers.js
```diff
@@ -49,3 +51,97 @@ export const customHeadingRenderer = (text, level, raw) => {
...
+ * @param {string} repoUrl
+ * @param {string} branchName
+ * @return {string}
+*/
+export const customLinkRenderer = (href, text, repoUrl, branchName) => {
+    // Ensure that anchors have lowercase href
+    if (href.startsWith('#')) {
+        href = href.toLowerCase();
+    }
+    // Only target relative URLs, which are used to refer to the git repo, and not anchors or absolute URLs
```
nguyeda1, line 121: Is this enough to check if href is relative URL?

m-murasovs: Yeah, you're right! Fix in

src/markdown_renderers.js
```diff
@@ -49,3 +51,98 @@ export const customHeadingRenderer = (text, level, raw) => {
...
+    const relativeUrlRegex = new RegExp('^(?:[a-z]+:)?//', 'i');
+    if (!relativeUrlRegex.test(href)) {
```
nguyeda1, line 123: I guess `relativeUrlRegex ` is regex for absolute URLs right? therefore it should be named `absoluteUrlRegex`.. also it looks like it fails if the URL is eg www.google.com see https://regex101.com/r/alX8HI/1

src/markdown_renderers.js
```diff
@@ -49,3 +51,98 @@ export const customHeadingRenderer = (text, level, raw) => {
...
+ * Mainly for use in actor READMES
+ * Parses the actor's repo URL to extract the repo name and owner name.
+ * @param {string} href
+ * @param {string} text
+ * @param {string} repoUrl
+ * @param {string} branchName
+ * @return {string}
+*/
+export const customImageRenderer = (href, text, repoUrl, gitBranchName) => {
+    if (!href.startsWith('http') && !href.startsWith('ftp') && !href.startsWith('#')) {
```
nguyeda1, line 142: This should be regex tested as well

nguyeda1 (changes requested):

nguyeda1 (changes requested):
One last thing otherwise looks good :) I would create a separate function `isUrlRelative` somewhere in utilities probably for checking the URL.. it might be useful in other usecases too :)

src/markdown_renderers.js
```diff
@@ -1,3 +1,7 @@
+import gitUrlParse from 'git-url-parse';
+
+const relativeUrlRegex = new RegExp('^(?!www.|(?:http|ftp)s?://|[A-Za-z]:\\|//).*', 'i');
```
jbartadev, line 3: Maybe put this to `src/regexs.js`

src/markdown_renderers.js
```diff
@@ -49,3 +53,98 @@ export const customHeadingRenderer = (text, level, raw) => {
             <h${level} id="${headingId}"><a href="#${headingId}"></a>${headingText}</h${level}>`;
     return headingToReturn;
 };
+
+export const parseRepoName = (repoUrl) => {
+    // Can't use parsedRepoUrl.full_name on it's own as Bitbucket adds irrelevant path suffix to the end of it
+    const parsedRepoUrl = gitUrlParse(repoUrl);
+    const repoNameParts = parsedRepoUrl.full_name.split('/');
+    const repoFullName = `${repoNameParts[0]}/${repoNameParts[1]}`;
+    return repoFullName;
```
jbartadev, line 62: Maybe something like this?
```
const parsedRepoUrl = gitUrlParse(repoUrl);
const repoName = parsedRepoUrl.full_name.split('/').slice(0,2).join('/');
return repoName;
```
The parse`RepoName` func name and the repo`FullName` return value don't match so it makes sense to do it like this. Or just return the result of the last command.
I don't really know how many parts are there in `repoNameParts` but probably still makes sense to use the slice and join...

src/markdown_renderers.js
```diff
@@ -49,3 +53,98 @@ export const customHeadingRenderer = (text, level, raw) => {
...
+    const repoFullName = parseRepoName(repoUrl);
+
+    const hrefParts = href.split('/');
+    const lastHrefPart = hrefParts[hrefParts.length - 1];
+
+    // If the last part of the URL has a dot, it's a file with an extension or .gitignore (blob),
+    // otherwise we assume the link is for a directory (tree)
+    const isTreeOrBlob = lastHrefPart.includes('.') ? 'blob' : 'tree';
+
+    if (repoUrl.includes('github.com')) {
```
jbartadev, line 97: It would be nice to have regexes for this too, probably not needed now

jbartadev (changes requested):
Looks good, just a minor comments

jbartadev (approved):
Looks good

nguyeda1 (approved):

PR #122, opened 2021-04-07
chore: actually deploy these files

vladfrangu:
Turns out, the current CD deploy script deploys just the files in the `build` directory ( :c )
This PR makes the types be next to the files (and yes, it works, I tested locally!)

CC: @pocesar @B4nan @mnmkng

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     },
     "homepage": "https://apify.com",
     "scripts": {
-        "build": "rm -rf build types && babel src --out-dir build && cp src/*.json build && tsc ./src/*.js --allowJs --declaration --emitDeclarationOnly --declarationDir types --skipLibCheck",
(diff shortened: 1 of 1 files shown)


pocesar (approved):
would just need to point the `types` / `typings` in package.json to this location I think

mnmkng:
Yeah, good point by @pocesar. Shouldn't we also change the `types` property of `package.json`? Now it points to the `types` directory which I assume would not exist after this change. @vladfrangu

vladfrangu:
> Yeah, good point by @pocesar. Shouldn't we also change the `types` property of `package.json`? Now it points to the `types` directory which I assume would not exist after this change. @vladfrangu 

Welll... If you look at what gets published, you'll see that the main and types entries are invalid no matter what. Realistically, both of those should be dropped from the JSON, or an index file should be created (where it exports everything as it's name). Which one of these sounds the best?

mnmkng:
@vladfrangu Yeah, I know that the `main` field is irrelevant. Noticed that when I was reviewing this PR. I don't know the impact of the `types` field though. But since you say it's ok, I'll merge it.

PR #125, opened 2021-04-09
Fix relative link rendering

m-murasovs:


src/markdown_renderers.js
```diff
@@ -114,9 +114,9 @@ export const generateGitRepoUrlPrefix = (repoUrl, branchName, href) => {
  * @return {string}
 */
 export const customLinkRenderer = (href, text, repoUrl, branchName) => {
-    // Handle anchor links and local Apify links
+    // Handle anchor links, local Apify links, and mailto
     // Return Apify domain links without rel="nofollow" for SEO
-    if (href.startsWith('#') || href.includes('apify.com')) {
+    if (href.startsWith('#') || href.includes('apify.com') || href.startsWith('mailto')) {
```
jbartadev, line 119: How about creating a regex for this? Already began to grow...
Thinking about something like `/^(mailto|tel|sms):.*$/` for identification that the href is this type of URI scheme so there won't be a need to change this again in the future.

jbartadev (changes requested):

src/markdown_renderers.js
m-murasovs: Yeah, good idea!

mnmkng (approved):

PR #128, opened 2021-04-14
Fix relative link render in READMEs

m-murasovs:
- Checking for repoUrl before rendering relative links - it was causing some errors
- Add fallback in case branch name is missing

This will make it so broken URLs don't cause errors


src/markdown_renderers.js
```diff
@@ -78,12 +80,15 @@ export const parseRepoName = (repoUrl) => {
  * Generates URLs for RAW content such as images
  *
  * @param {string} repoUrl
- * @param {string} branchName
+ * @param {string} gitBranchName
```
jbartadev, line 83: Maybe when you are renaming this, could you also rename the `repoUrl`too to have this consistent?

src/markdown_renderers.js
```diff
@@ -78,12 +80,15 @@ export const parseRepoName = (repoUrl) => {
...
  */
-export const generateRawGitRepoUrlPrefix = (repoUrl, branchName) => {
+export const generateRawGitRepoUrlPrefix = (repoUrl, gitBranchName) => {
     let urlPrefix;
     const repoFullName = parseRepoName(repoUrl);
 
+    // Avoid errors created by missing branch name / badly formed URLs
+    const branchName = gitBranchName || 'main';
```
jbartadev, line 90: The `'main'` looks like a constant

src/markdown_renderers.js
```diff
@@ -99,10 +104,10 @@ export const generateRawGitRepoUrlPrefix = (repoUrl, branchName) => {
  * Generates URLs for files and folders
  *
  * @param {string} repoUrl
```
jbartadev, line 106: The same as above

jbartadev (changes requested):
OK, just a small code style comments

jbartadev (approved):

PR #129, opened 2021-04-15
Feature: makeInputJsFieldsReadable with option to return the functions as template literals

nguyeda1:
Current actor example codes that are passing functions in the input are not copy/paste/runnable. Tested on https://apify.com/apify/web-scraper/api 

The function needs to be passed as a string to the client in order for it to work (makes it uglier in the UI but functional). 
### Before
Screen Shot 2021-04-15 at 15 32 42

### After
Screen Shot 2021-04-15 at 15 33 21

**NOTE**:
The fact that no one has reported and issue with this (AFAIK) could mean that no one actually use these


jancurn:
@nguyeda1 AFAIK the code example worked fine, since the `Apify.call()` function automatically translated the functions to string. Maybe that's why nobody complained. Or what exactly didn't work for you?

nguyeda1:
> @nguyeda1 AFAIK the code example worked fine, since the `Apify.call()` function automatically translated the functions to string. Maybe that's why nobody complained. Or what exactly didn't work for you?

Oh right maybe it is not working for the ApifyClient only then. I was about to rewrite the example to use the new client and @mnmkng suggested to skip the `apify` package in favor of `apify-client`. The ApifyClient did not accept the pageFunction as a valid input unless it was a string.

mnmkng:
Oh, snap. I did not know that there are some magic JSON parsers in SDK `Apify.call()`. We would have to add this to the client as well. But why not, stringifying functions to JSON makes sense to me.

jancurn:
Yeah that would be cool, I think we have some magic code in SDK for that which can be reused.

nguyeda1:
Closing this since we are going to add the serialization to the client directly

PR #132, opened 2021-04-25
Hubspot client

gippy:
Implements the necessary functions for hubspot integration from app and lead generation from web.
The hubspot api is accessed through this client: https://github.com/HubSpot/hubspot-api-nodejs
Docs for the api are here: https://developers.hubspot.com/docs/api/overview

The API key in unit tests is not old API token from testing account. After I finished the PR I created new token under the test account and disabled the old one.


src/hubspot_client.js
```diff
@@ -0,0 +1,629 @@
...
+ * @return {Object} Object with only keys allowed in the schema
+ */
+export function cleanAndCompareWithSchema(data, allowedValues) {
+    const cleanedData = {};
+    Object.keys(allowedValues).forEach((key) => {
+        // Skip schema fields that are not present in data
+        if (typeof data[key] === 'undefined') return;
+        // Save values set to null without typechecking
+        if (data[key] === null) {
+            // cleanedData[key] = null;
```
novotnyj, line 82: This is commented out on purpose? :)

novotnyj (approved):

src/hubspot_client.js
gippy: Actualy this is copyied from current salesforce client without any changes. Reason is probably that salesforce did not accept null values, looking at hubspot it does not either, so I guess we can change this to empty string or something if it causes issues.

PR #137, opened 2021-05-04
refactor: split into multiple packages + TS rewrite

### LINKED ISSUE #95: deprecated request@2.88.2
npm WARN deprecated request@2.88.2: request has been deprecated

image

### LINKED ISSUE #131: Monorepo rewrite
As @mnmkng mentioned, would be good to split this into multiple packages and convert the repo to monorepo, so we do not need to use deep imports ([which is in general a bad practice](https://gist.github.com/daleyjem/0f38f561a4e91e58eba580889f38330f#why-deep-imports-can-be-a-problem)). It will also serve as a PoC for future monorepos like the SDK. 

## Proposal

There are several options, mostly dependent on package manager (e.g. we could use `pnpm` or `yarn` that truly support workspaces), but in the end we will still probably need `lerna` too. It will help mainly with release management and build orchestration. 

Here is the proposed set of repositories (partially based on usage, partially on 3rd party dependencies):

> Unfortunately we can't use package names like `apify-shared/consts` (forbidden by NPM), so it will have to be something like `@apify-shared/consts`.

- `@apify-shared/consts`
	- `consts.js`
	- `regexs.js`
- `@apify-shared/log`
	- `log.js`
	- `log_consts.js`
	- `log_helpers.js`
	- `logger.js`
	- `logger_json.js`
	- `logger_text.js`
- `@apify-shared/input_schema`
	- `input_schema.js`
	- `input_schema.json`
	- `intl.js`
- `@apify-shared/git`
	- `git.js`
- `@apify-shared/salesforce_client`
	- `salesforce_client.js`
- `@apify-shared/datastructures`
	- `linked_list.js`
	- `list_dictionary.js`
	- `lru_cache.js`
- `@apify-shared/markdown`
	- `markdown_renderers.js`
	- `marked.js`
- `@apify-shared/utilities`
	- `exponential_backoff.js`
	- `health_checker.js`
	- `image_proxy_client.js`
	- `parse_jsonl_stream.js`
	- `streams_utilities.js`
	- `utilities.client.js`
	- `utilities.js`
	- `webhook_payload_template.js`

Happy to hear suggestions about this, as I haven't seen much of the internal repos so this is mostly based on the SDK. This will have direct impact on breaking changes, so if there is a file that is worth having as a single file repository just because it is used extensibly, we should indeed do it that way, especially if its in client facing code. 

We might also want to split some of the large files (e.g. the `consts.js` one is quite large).

### TypeScript

Personally I would also use this opportunity to convert the repository to typescript. We already added automatic generation of types recently, but they are all based purely on JS inference, so not great in general. Also I do have a [working setup of lerna with TS](https://github.com/mikro-orm/mikro-orm/pull/475) so it would actually be easier for me to set it up that way. 

### Removal of default exports

Another thing I don't like is the overusage of default exports. They just add complexity to the compiled code, and also reduces DX (find usage & go to declarations are buggy as they jump to the import line instead of the actual implementation, at least in WS). If we are fine with making some additional breaking changes, I'd suggest getting rid of them in most places (haven't checked much code, I could understand the one in `log`, if we need to export an instance, default export sounds like the right thing to do). 

### Removal of polyfills

The only present polyfill is `startsWith` - I believe its already the time when its safe to say ES6 is supported everywhere, so this should not be needed. Unless we have actual need for this, let's get rid of it. 

### Removal of `underscore`

It's the ES6 time, most of the helper methods are natively supported or easily implemented on very few LoC. Mostly used ones that are clearly not needed: `isArray`, `mapObject`, `map`, `some`, `filter`, `extends` (as per the docs its the same as `Object.assign()`, it's not recursive), also pretty much all of the `is*` methods can be replaced with single line variants (either `typeof` or `instanceof`). I'd probably keep it in the salesforce client package, as there it is used heavily, but most of the other usages are easy to replace. 

Btw if you think underscore is a small dependency, it's not (also native methods will be always faster):

```sh
➜  apify-shared-js git:(master) ✗ npx howfat underscore
Dependencies: 0
Size: 880.22kb
Files: 503
```

### Yarn workspaces

Personally a big fan of `yarn` (v1), using it for years and never did me wrong, and it makes a good tandem when combined with `lerna`. NPM recently added support for workspaces too, but from what I read, it's far from complete. Here is some reading about why it is good combination:

https://doppelmutzi.github.io/monorepo-lerna-yarn-workspaces/

With `lerna` we should be fine with NPM too (as it was supported way even before NPM added workspaces), so this is really just an open question. But as I mentioned above, I do have a working version of this and it also uses yarn workspaces, so doing it this way would ease the implementation for me (or better - remove some unknowns - it could be also as easy as with yarn).

Also worth reading: https://hyperfoo.io/posts/npm-7-workspaces-1

## Alternative

We might also just add `index.js` file and reexport things from there, and instead of accessing `apify-shared/consts` we could do `import { consts } from 'apify-shared'`.

B4nan:
Packages:
- `@apify/consts`
- `@apify/datastructures`
- `@apify/git`
- `@apify/hubspot_client`
- `@apify/image_proxy_client`
- `@apify/input_schema`
- `@apify/log`
- `@apify/markdown`
- `@apify/salesforce_client`
- `@apify/utilities`

Other notable changes:
- some of the typings are rather low effort (lots of `any` in some places, mainly salesforce and hubspot clients)
- uses named exports except `log` where we keep the default export for logger instance
- removes `underscore` from most of the packages (except salesforce/hubspot clients)
- removed `startsWith` polyfill and `newPromise` and `requestPromised` methods
- `truncate` method is duplicated so the `log` package does not depend on `utilities` (as those bring a lot of 3rd party deps)

Closes #131
Closes #95

BREAKING CHANGE:
- old `apify-shared` package is now gone in favour of new `@apify/*` packages
- all exports are now done via named exports instead of default exports (with exception of logger instance)
- removed `startsWith` polyfill and `newPromise` and `requestPromised` methods


mnmkng:
@mtrunkat @fnesveda I think it's important now to finally agree on the structure and principles behind this PR before we do a full code review. We haven't really done so in the issue.

I'm ok with the way it's done.

jancurn:
> @mtrunkat @fnesveda I think it's important now to finally agree on the structure and principles behind this PR before we do a full code review. We haven't really done so in the connected issue.
> 
> I'm ok with the way it's done.

I guess this decision also depends on the overall strategy for the naming of our open source NPM packages. For example, to say we're going to use `@apify/xxx` mainly internal tools and packages, and generic names for public reusable ones. But that means the "apify" brand will not be present in the generic public packages, which might or might not be cool.

What do you think?

B4nan:
So I managed to make the tests work with node < 15 too, and have some concrete points on what is needed for what. 

- Current setup uses NPM without workspaces, linking packages via lerna. It works well enough, but it requires us to have lock files (not just for the root project, but for all packages too), to be able to install the project in CI (don't ask me why, afaik its a clash between lerna and NPM, as lerna is fiddling with the lock files too).
- If we use yarn (v1), we should be probably fine without lockfiles (it would definitely work with just the root level one - had that setup working elsewhere - and probably even without that one too).
- If we use NPM workspaces, it works only on node 15+ (even if we manually upgrade NPM, again don't ask me why 🤷).
- There is one small gotcha with the way it is now (NPM without workspaces) - if we want to use a dependency of a package in tests, we need to explicitly include it in the main project dev deps (with workspaces this works automatically).

Will focus on the linting and publishing setup now. Also planning to make the readme much more verbose, adding info about how to add new package, where to put dev deps, etc.

mnmkng:
@jancurn We already had this discussion and figured that generic packages don't really take off unless someone spends non-trivial time marketing them. And since we neither want to market the packages nor keep up with potential community feature requests, we decided to just place them all under `@apify`. But that concerns only `apify-shared` we still want to release `puppeteer-scraper` `cheerio-scraper` and so on under the hood of the SDK.

packages/consts/src/consts.ts
```diff
@@ -376,6 +376,7 @@ export const ENV_VARS = {
     FACT: 'APIFY_FACT',
     DEDICATED_CPUS: 'APIFY_DEDICATED_CPUS',
     SDK_LATEST_VERSION: 'APIFY_SDK_LATEST_VERSION',
+    INTERNAL_PORT: 'APIFY_INTERNAL_PORT',
```
mnmkng, line 379: I have absolutely no idea where this is coming from. Why add it?

mnmkng (changes requested):
I did not really read the code, because the tests are passing and it would take me a week. But I don't think that's a problem, because the packages are completely new, so a manual migration is needed anyway. Just commented on some nitpicks and typos.

packages/consts/src/consts.ts
B4nan: it is used just few lines below this, we need to define all constats that are used in the code, otherwise TS will complain. 

(or we could typecast, which is rather ugly in this case, as we'd have to do `(ENV_VARS as any).INTERNAL_PORT`)

.github/workflows/test_and_release.yml
```diff
@@ -0,0 +1,121 @@
...
+        steps:
+            -   uses: actions/checkout@v2
+                with:
+                    token: ${{ secrets.GH_TOKEN }}
+                    fetch-depth: 0 # we need to pull everything to allow lerna to detect what packages changed
+                    ref: master
+            -   uses: actions/setup-node@v1
+                with:
+                    node-version: 16
+                    registry-url: https://registry.npmjs.org/
```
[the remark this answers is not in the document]
B4nan: apparently not, also switched to v2 of setup-node

packages/consts/src/consts.ts
B4nan: actually looking at the original code, and apparently this was never working, as the `ENV_VARS.INTERNAL_PORT` was not defined, so the usage in `INTEGER_ENV_VARS` resulted in `undefined`

so what to do with this, remove both usages I guess?

mnmkng: Yeah, I think we can delete it. I've never seen it used anywhere.

B4nan:
Looking at the `utilities` package, as it is also used in SDK and now it still contains way too many 3rd party dependencies. Some ideas:

- there is the deprecated `request` package used for a single method - `requestPromised`, I'd propose to get rid of that, if we use it somewhere, we should stop anyway as it is deprecated and we now have many good replacements
- `ImageProxyClient` has an external dependency on `create-hmac` (which itself has [5 transitive deps](https://github.com/crypto-browserify/createHmac/blob/master/package.json#L36), havent checked deeper), so it could go to its own package
- `validateProxyField` method uses `countries-list` which is also huge, so worth separating (maybe there should be other methods moved to such new package?)
- `validateInputUsingValidator` depends on `ajv`, maybe that one belongs more to `@apify/input_schema` package (where we depends on `ajv`)?
- `traverseObject` depends on `is-buffer`, node already have `Buffer.isBuffer()` and that package is basically a one liner using it (although in a bit obfuscated way, e.g. `obj.constructor.isBuffer === 'function' && obj.constructor.isBuffer(obj)`)

Thoughts?

mnmkng:
`request` - yes, remove it
`ImageProxyClient` - makes sense to me to separate it
`validateProxyField` - is this releated to the proxy client above?
`validateInputUsingValidator` - not sure about this one, please check `apify-cli`, I have a hunch it might be used there
`traverseObject` - I think that check is there for browser compatibility, we can copy paste it and drop the dependency

B4nan:
There are more candidates to be moved to `input_schema` package. All of those are dependent on that package, if we move them all, we would get rid of the dependency `utilities -> input_schema` (which again has some transitive deps)

- `validateProxyField`
- `validateInputUsingValidator`

(`validateProxyField` is where the `countries-list` is used, so that would go to `input_schema` dependencies, making the `utilities` package depend only on `log` and `consts` packages)

> validateProxyField - is this releated to the proxy client above?

Not sure, probably not. It's a separate method, it is used only in `validateInputUsingValidator` in the repo.

> validateInputUsingValidator - not sure about this one, please check apify-cli, I have a hunch it might be used there

Does it matter? We are talking about moving the code to different package, it will still be somewhere.

> traverseObject - I think that check is there for browser compatibility, we can copy paste it and drop the dependency

Ok will inline it then, at least for now.

B4nan:
There is also `makeInputJsFieldsReadable` that adds `cherow` as dependency (which is JS parser, by the way a bit outdated with more recent replacements from the very same author, namely there is `escaya` that supports TS too).

Sounds like another things that should go to the `input_schema` package.

mnmkng:
> Does it matter? We are talking about moving the code to different package, it will still be somewhere.

I'm not saying we should not do it. I'm just saying we should have an idea about the usage first. But yeah, `cli` is pretty easy to update (if we actually do update it at some point).

It would be worth it to pull at least the `apify-core` repo and do a search for all those methods there, to have an idea about how widespread their usage is and how painful the migration from `apify-shared` to `@apify/...` would be.

gippy:
The `validateInputUsingValidator` is used in app to correctly validate input in input UI and are required in `storage`. Should be no problem to import in from somewhere else then apify-shared.
The `validateProxyField` is probably only used in the `validateInputUsingValidator` function.

B4nan:
Looking at the `create-hmac` dependency, why don't we use native `crypto` package instead?

https://nodejs.org/api/crypto.html#crypto_class_hmac

It was like that originally, but changed in favour of the 3rd party package here:

https://github.com/apify/apify-shared-js/commit/0d4d0241b71490e98db69ecedfe89955b470a3f5

But no reasoning behind it in the commit message 😢 Apparently the 3rd party one is browser ready? Maybe we should have a separate package for browser code so we have clear separation of what can use node API's and what not?

B4nan:
Lol, `cherow` is really FAT :D 

```
➜  apify-shared-js git:(monorepo) npx howfat cherow
Dependencies: 1270
Size: 56.69mb
Files: 15682
```

(yes, of course we have that pulled in everywhere, like in the SDK and all its dependents)

B4nan:
Managed to replace `cherow` with `escaya`, tests are passing. Will still make it a separate package, same for `ImageProxyClient`. 

Or maybe the `makeInputJsFieldsReadable` could go into `input_schema` package, WDYT?

fnesveda:
> Looking at the `create-hmac` dependency, why don't we use native `crypto` package instead?
> 
> [nodejs.org/api/crypto.html#crypto_class_hmac](https://nodejs.org/api/crypto.html#crypto_class_hmac)
> 
> It was like that originally, but changed in favour of the 3rd party package here:
> 
> [0d4d024](https://github.com/apify/apify-shared-js/commit/0d4d0241b71490e98db69ecedfe89955b470a3f5)
> 
> But no reasoning behind it in the commit message 😢 Apparently the 3rd party one is browser ready? Maybe we should have a separate package for browser code so we have clear separation of what can use node API's and what not?

Yeah, I switched it from `crypto` to `create-hmac` so that we don't have to polyfill the whole `crypto` on browser as a part of an effort to reduce the size of the web bundle. The reason was included in the PR description (https://github.com/apify/apify-shared-js/pull/87) and in the changelog, but probably didn't get copied to the commit message on merge.

mnmkng:
> Or maybe the makeInputJsFieldsReadable could go into input_schema package, WDYT?

I checked `apify-core` and `apify-web` and it's used mostly standalone, not in connection with input schema so I would not put it there.

mtrunkat:
- request - yes, please remove. We have already replaced request with Axios hopefully everywhere and if we missed some plases then we should do it anyway.
- ImageProxyClient - We could move this to internal package
- traverseObject - As I am looking into the code I have feeling that all functions that use this are now only used at the backend so we could potentially get a rid of the dependency. The reason why it's in a client side module was a legacy crawler (my guess). But I would have to dig into  this a bit more.

These are input schema related (although not directly connected) so I may put everything to input_schema for simplicity:
- validateProxyField
- validateInputUsingValidator
- makeInputJsFieldsReadable

@B4nan good job with cheerow!

B4nan:
Ok good, moved `ImageProxyClient` to separate package (`@apify/image_proxy_client`, will be happy to hear suggestions for better name), removed `request` and moved the `makeInputJsFieldsReadable` to `input_schema` package. Regarding the `is-buffer`, as it was just a two line method, it is now inlined to the utilities, so it's fine to keep the `traverseObject` there for now. 

Also replaced `chalk` (6 deps) with `ansi-colors` (0 deps). The API is compatible, with few minor differences (e.g. the tagged template function is missing, but personally I think that is the worst way to use colors in CLI, as there are no type checks/intellisense).

```
➜  apify-shared-js git:(monorepo) ✗ npx howfat chalk
Dependencies: 5
Size: 94.86kb
Files: 33
➜  apify-shared-js git:(monorepo) ✗ npx howfat ansi-colors
Dependencies: 0
Size: 23.82kb
Files: 6
```

https://github.com/doowb/ansi-colors

Overall it should be ready for merging. 

One thing to note - we should mark the commit as breaking so the release will be 1.0. Will adjust the PR description that way, so just a head's up that when merging, the PR description should be copied to the commit message (I guess it will be better if I will merge it myself so I can cancel the build if I see something wrong in the publishing step).

mnmkng (approved):

B4nan:
Added summary of what changed with regards to the monorepo redesign and automatic publishing into the repo wiki:

https://github.com/apify/apify-shared-js/wiki/The-monorepo-rewrite-story

mnmkng:
@B4nan Can I merge this?

B4nan:
We need to add the personal access token to secrets first (with write permissions as it will be used for commiting to the repo). Should be called `GH_TOKEN`.

Ideally we should use some service account for that, I think I saw one recently in one of the repositories.

PR #173, opened 2021-06-02
fix: make `configureLogger` accept `Log` instance instead of `Logger`

B4nan:
Related: https://github.com/apify/apify-conductor/pull/103

### DIFF
--- a/packages/utilities/src/utilities.ts
+++ b/packages/utilities/src/utilities.ts
@@ -8,7 +8,7 @@
  */
 
 import crypto from 'crypto';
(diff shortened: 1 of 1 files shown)


mnmkng (approved):

mnmkng:
Let's see what happens 😄

B4nan:
damn, failed to ship to NPM, and I kinda know why, the versions are taken from `dist/package.json` that is outdated, only the root one has correct  version. will ship it manually now and try to deal with this tomorrow.

B4nan:
publishing should be hopefully fixed, let's see on the next PR, there is already one waiting for review

PR #174, opened 2021-06-02
feat: Increased length of actor name

drobnikj:
Increased the length of actor name into 63 chars (DNS safe string length)

### DIFF
--- a/packages/consts/src/consts.ts
+++ b/packages/consts/src/consts.ts
@@ -234,7 +234,7 @@ export const USERNAME = {
  */
 export const ACTOR_NAME = {
     MIN_LENGTH: 3,
-    MAX_LENGTH: 30,
+    MAX_LENGTH: 120,
     REGEX: DNS_SAFE_NAME_REGEX,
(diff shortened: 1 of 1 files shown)


mtrunkat (approved):

mnmkng:
I don't understand the problem much, but why would an actor name, that's in the path part of URL need to be only 63 chars? Doesn't everyone use longer strings?

mnmkng:
e.g. GitHub uses 100 chars for repository names

drobnikj:
We discussed it with Mara and decided to go with DNS safe string to do not break in the future.
cc: https://github.com/apify/apify-core/issues/2792#issuecomment-853801044

mnmkng:
Cool, will merge it when I get a confirmation from @B4nan that the deployment is fixed.

B4nan:
Only way to find out is to merge it :D

B4nan:
Works!

https://github.com/apify/apify-shared-js/releases

PR #231, opened 2021-07-13
fix: replace git-url-parse with native method

m-murasovs:
Replacing the `git-url-parse` package in the markdown renderers with native Node URL parsing (it's not compatible with Safari).

### DIFF
--- a/packages/markdown/src/markdown_renderers.ts
+++ b/packages/markdown/src/markdown_renderers.ts
@@ -50,7 +50,7 @@ export function customHeadingRenderer(text: string, level: 1 | 2 | 3 | 4 | 5 | 6
     return `\n${' '.repeat(12)}<h${level} id="${headingId}"><a href="#${headingId}"></a>${headingText}</h${level}>`;
 }
 
-export function getRepoFullName(gitRepoUrl: string): string {
+export function parseRepoName(gitRepoUrl: string): string {
     const parsedRepoUrl = new URL(gitRepoUrl);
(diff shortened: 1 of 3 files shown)


packages/markdown/src/markdown_renderers.ts
```diff
@@ -51,18 +50,19 @@ export function customHeadingRenderer(text: string, level: 1 | 2 | 3 | 4 | 5 | 6
     return `\n${' '.repeat(12)}<h${level} id="${headingId}"><a href="#${headingId}"></a>${headingText}</h${level}>`;
 }
 
-export function parseRepoName(gitRepoUrl: string): string {
-    const parsedRepoUrl = gitUrlParse(gitRepoUrl);
+export function getRepoFullName(gitRepoUrl: string): string {
```
jbartadev, line 53: This is a breaking change, please always try to not do breaking changes for a public package like this one when you don't have to. I know that this is a better naming but if the returned value is still the same there is no reason to break it for everyone who potentially uses this function.

jbartadev (changes requested):
This looks like an easy replacement. Small comments.

 The `git-url-parse` package isn't used anymore in this package so please remove it from dependency too.

packages/markdown/src/markdown_renderers.ts
m-murasovs: Good catch - I totally didn't consider it!

jbartadev (approved):
Looks good.

PR #233, opened 2021-07-14
feat: Added FREE_MAIL_PROVIDERS to consts

novotnyj:
Will replace [this](https://github.com/apify/apify-web/blob/962ffaab29f349681c7b5f50fb44d4942266f763/src/ui/custom_components/CustomProjectForm/customProjectForm.consts.ts) and will be used in app to decide whether to call cognism api or not...

### DIFF
--- a/packages/consts/src/consts.ts
+++ b/packages/consts/src/consts.ts
@@ -604,3 +604,18 @@ export const MARKETPLACE_USER_ROLES = {
 };
 
 export const GIT_MAIN_BRANCH = 'main';
+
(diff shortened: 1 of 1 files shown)


nguyeda1 (approved):

mtrunkat (changes requested):
It would be better to put this into costs in `apifier-storage` as this does not need to be sharted with our tooling just internal apps and services

novotnyj:
Makes sense, closing this in favour of `apifier-storage`

PR #237, opened 2021-07-19
fix: include SSH URL handling in repo name parser

m-murasovs:
This fixes the replacement of the git-url-parse package, allowing the parsing of SSH URLs


packages/markdown/src/markdown_renderers.ts
```diff
@@ -51,8 +51,14 @@ export function customHeadingRenderer(text: string, level: 1 | 2 | 3 | 4 | 5 | 6
 }
 
 export function parseRepoName(gitRepoUrl: string): string {
-    const parsedRepoUrl = new URL(gitRepoUrl);
-    const path = parsedRepoUrl.pathname.substr(1);
+    // Handling for SSH URLs
```
B4nan, line 54: let's add some tests for this so we don't break it again (`./test/markdown_renderers.test.ts`)

packages/markdown/src/markdown_renderers.ts
```diff
@@ -51,8 +51,14 @@ export function customHeadingRenderer(text: string, level: 1 | 2 | 3 | 4 | 5 | 6
...
+    const normalizedUrl = gitRepoUrl.includes('git@')
+        ? gitRepoUrl.replace(':', '/').replace('git@', 'https://')
+        : gitRepoUrl;
+    const parsedRepoUrl = new URL(normalizedUrl);
+    const cleanedPath = parsedRepoUrl.pathname.replace('.git', '');
+    // Do not use the initial slash in the path
+    const path = cleanedPath.substr(1);
     // Can't use parsedRepoUrl.full_name on it's own as Bitbucket adds irrelevant path suffix to the end of it
```
B4nan, line 62: this comments is probably not relevant anymore? or maybe it should be just rephrased a bit, as now there is no `parsedRepoUrl.full_name`.

B4nan (changes requested):

B4nan (approved):

PR #238, opened 2021-07-20
chore: improve markdown tests and inputs

m-murasovs:
I noticed that the tests could have nicer input data, so this is the update. No need for release, it's just for my own peace of mind :) 

also adding a test for image like generation with SSH URLs

### DIFF
--- a/packages/markdown/src/markdown_renderers.ts
+++ b/packages/markdown/src/markdown_renderers.ts
@@ -159,5 +159,5 @@ export function customImageRenderer(href: string, text: string, gitRepoUrl: stri
         href = `${urlPrefix}/${cleanedHref}`;
     }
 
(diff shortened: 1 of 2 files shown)


B4nan (approved):

B4nan:
it should not produce a release when the pr title is marked as `chore`, let's see in a minute if it works correctly :]

B4nan:
mmm it does not work as expected (new version of the markdown package was released), weird, will need to review what's happening. but an extra patch release won't hurt anything i guess 🤷 apparently it works for the `build` type and not for `chore` as my `build` commits did not produce a release, but i also pushed those directly to master and not via pr...

m-murasovs:
weird indeed 🤔 but yeah, it's not a problem if there's an extra release 🙂

PR #244, opened 2021-08-16
feat: add APIFY_API_PUBLIC_BASE_URL

dragonraid:
Adding `APIFY_API_PUBLIC_BASE_URL` that will contain public apify cloud's url.


B4nan (approved):
so this is the one we should let people configure in the SDK, right? 

edit: or better say the one that can be changed based on local/dev/prod environments

dragonraid:
> so this is the one we should let people configure in the SDK, right?
> 
> edit: or better say the one that can be changed based on local/dev/prod environments

This is the public URL of API. With this url customers would be able to access API outside of Apify cloud, unlike `API_BASE_URL`, which is private.

Does that answer the question?

PR #247, opened 2021-08-23
feat: update input schema to use ajv v8

valekjo:
BREAKING CHANGE: update ajv to v8 and set it as peerDependency

### DIFF
--- a/packages/input_schema/package.json
+++ b/packages/input_schema/package.json
@@ -36,8 +36,10 @@
     },
     "dependencies": {
         "@apify/consts": "^1.2.0",
-        "ajv": "^6.12.6",
         "countries-list": "^2.6.1",
         "escaya": "^0.0.61"
+    },
+    "peerDependencies": {
+        "ajv": "^8.6.2"
     }
 }
--- a/packages/input_schema/package.json
+++ b/packages/input_schema/package.json
@@ -40,6 +40,6 @@
         "escaya": "^0.0.61"
     },
     "peerDependencies": {
-        "ajv": "^8.6.2"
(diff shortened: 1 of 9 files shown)


packages/input_schema/src/input_schema.ts
```diff
@@ -116,6 +116,7 @@ const validateField = <T extends Record<string, any>> (validator: Ajv, fieldSche
  * This function validates given input schema first just for basic structure then each field one by one and
  * finally fully against the whole schema.
  *
+                    q: {},
```
B4nan, line 119: i guess this is not intentional?

packages/input_schema/package.json
```diff
@@ -36,8 +36,10 @@
     },
     "dependencies": {
         "@apify/consts": "^1.2.0",
-        "ajv": "^6.12.6",
         "countries-list": "^2.6.1",
         "escaya": "^0.0.61"
+    },
+    "peerDependencies": {
+        "ajv": "^8.6.2"
```
B4nan, line 43: peer deps are usually just about major versions, so as long as we know there are no major issues with 8.0.0, i'd suggest to use that instead

```suggestion
        "ajv": "^8.0.0"
```

B4nan (approved):
looking good, just some nitpicks

btw the `BREAKING CHANGE` in the commit message is with space, not underscore. when merging, ensure the commit message contains it - otherwise it won't be marked as BC and will be just a minor version bump (it is about the resulting commit message in master, not about PR description, i know gh sometimes don't copy it automatically to the field).

PR #249, opened 2021-08-25
feat: Add APIFY_DISABLE_OUTDATED_WARNING to available env vars in consts

fnesveda:


mnmkng:
Why do we need this in shared? It's only used in the SDK afaik.

PR #250, opened 2021-08-26
feat: add customer_segment to hubspot_client

HonzaKirchner:


HonzaKirchner:
CleanShot 2021-09-14 at 10 16 26@2x

PR #267, opened 2021-11-19
feat: add `@apify/timeout` package with abortable promise timeout helper

B4nan:
Runs given handler and rejects with the given `errorMessage` (or Error instance) after given `timeoutMillis`, unless the original promise resolves or rejects earlier. Use `tryCancel()` function inside the handler after each await to finish its execution early when the timeout appears.

```ts
import { addTimeoutToPromise, tryCancel } from '@apify/timeout';

async function doSomething() {
    await doSomethingTimeConsuming();
    tryCancel();
    await doSomethingElse();
    tryCancel();
}

const res = await addTimeoutToPromise(
  () => handler(),
  200,
  'Handler timed out after 200ms!',
);
```

This uses `AsyncLocalStorage` to hold the `AbortController` instances created inside the `addTimeoutToPromise` function. The context is shared for this package, and respects nesting of `addTimeoutToPromise` calls. This means it can be used in different packages to cancel the same nested handler, e.g. we can create the timeout in SDK and use `tryCancel` from inside browser pool (we can still handle the timeouts explicitly in browser pool, and it will automatically used the upper context if there is some0.

### DIFF
--- a/packages/timeout/src/index.ts
+++ b/packages/timeout/src/index.ts
@@ -1,3 +1,4 @@
+// eslint-disable-next-line max-classes-per-file
 import { setTimeout } from 'timers/promises';
 import { AsyncLocalStorage } from 'async_hooks';
 
@@ -14,10 +15,16 @@ export interface AbortContext {
 export const storage = new AsyncLocalStorage<AbortContext>();
 
 /**
+ * Custom error class that will be used for timeout error.
+ */
+export class TimeoutError extends Error {
+}
+
+/**
  * Custom error class to handle `tryCancel()` checks.
  * This should not be exposed to user land, as it will be caught in.
  */
-export class TimeoutError extends Error {
+class InternalTimeoutError extends TimeoutError {
 }
 
 /**
@@ -40,7 +47,7 @@ export function tryCancel(): void {
     const { signal } = storage.getStore()?.cancelTask ?? {};
 
     if (signal?.aborted) {
-        throw new TimeoutError('Promise handler has been canceled due to a timeout');
+        throw new InternalTimeoutError('Promise handler has been canceled due to a timeout');
     }
 }
 
@@ -59,8 +66,11 @@ export function tryCancel(): void {
  * ```
  */
 export async function addTimeoutToPromise<T>(handler: () => Promise<T>, timeoutMillis: number, errorMessage: string | Error): Promise<T> {
-    const cancelTimeout = new AbortController();
-    const cancelTask = new AbortController();
+    // respect existing context to support nesting
+    const context = storage.getStore() ?? {
+        cancelTimeout: new AbortController(),
+        cancelTask: new AbortController(),
+    };
     let ret: T;
 
     // calls handler, skips internal `TimeoutError`s that might have been thrown
@@ -69,20 +79,18 @@ export async function addTimeoutToPromise<T>(handler: () => Promise<T>, timeoutM
         try {
             ret = await handler();
         } catch (e) {
-            if (!(e instanceof TimeoutError)) {
+            if (!(e instanceof InternalTimeoutError)) {
                 throw e;
             }
         } finally {
-            storage.getStore()?.cancelTimeout.abort();
+            context.cancelTimeout.abort();
         }
     };
 
     const timeout = async () => {
-        const store = storage.getStore();
-
         try {
-            await setTimeout(timeoutMillis, undefined, { signal: store?.cancelTimeout.signal });
-            store?.cancelTask.abort();
+            await setTimeout(timeoutMillis, undefined, { signal: context.cancelTimeout.signal });
+            context.cancelTask.abort();
         } catch (e) {
             // ignore rejections (task finished and timeout promise has been cancelled)
             return;
@@ -94,7 +102,7 @@ export async function addTimeoutToPromise<T>(handler: () => Promise<T>, timeoutM
     };
 
     await new Promise((resolve, reject) => {
-        storage.run({ cancelTimeout, cancelTask }, () => {
+        storage.run(context, () => {
             Promise.race([timeout(), wrap()]).then(resolve, reject);
         });
     });
--- a/packages/timeout/src/index.ts
+++ b/packages/timeout/src/index.ts
@@ -44,7 +44,7 @@ class InternalTimeoutError extends TimeoutError {
  * ```
  */
 export function tryCancel(): void {
-    const { signal } = storage.getStore()?.cancelTask ?? {};
+    const signal = storage.getStore()?.cancelTask.signal;
 
     if (signal?.aborted) {
         throw new InternalTimeoutError('Promise handler has been canceled due to a timeout');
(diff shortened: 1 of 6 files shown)


szmarczak:
Do we need `AsyncLocalStorage` at all? I've tried

```js
export async function addTimeoutToPromise(handler, timee, errorMessage) {
    const cancelTimeout = new AbortController();

    return Promise.race([
        (async () => {
            try {
                await setTimeoutAsync(timee, undefined, { signal: cancelTimeout.signal });
            } catch {
                return;
            }

            throw new Error(errorMessage);
        })(),
        handler.finally(() => {
            cancelTimeout.abort();
        }),
    ]);
}
```

and it works as well.

B4nan:
And how are you going to implement the early escape via `tryCancel()`? That is why we need ALS, so we don't have to maintain the `AbortController` instances (and we automatically handle parent contexts, even cross package ones, which is needed for our use case).

szmarczak:
We could pass `tryCancel` as an argument to `handler`. But yeah, I can see `AsyncLocalStorage` being more feasible :+1:

szmarczak (approved):

szmarczak:
> we need to escape the handler as early as possible to avoid sideeffects (like those double processed requests).

There is still a chance that it will time out when sending the actual HTTP request (when updating the storage). I think that it would be better to pass `signal` instead.

B4nan:
> There is still a chance that it will time out when sending the actual HTTP request (when updating the storage). I think that it would be better to pass signal instead.

We still can (once got supports this, it's not there currently, right?). The signal is exposed via `storage.getStore()?.cancelTask.signal` (we can add a shortcut for that if we need it). But what I am trying to solve now is not about http timeouts.

szmarczak:
Ah, right. In this case it would look like

```js
if (signal) {
	gotPromise.cancel();
	gotPromise.catch(() => {});
}
```

LGTM.

packages/timeout/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+ * );
+ * ```
+ */
+export async function addTimeoutToPromise<T>(handler: () => Promise<T>, timeoutMillis: number, errorMessage: string | Error): Promise<T> {
+    // respect existing context to support nesting
+    const context = storage.getStore() ?? {
+        cancelTimeout: new AbortController(),
+        cancelTask: new AbortController(),
+    };
+    let ret: T;
```
mnmkng, line 74: Would it hurt to name it `returnValue` ? 😄

packages/timeout/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+ * async function doSomething() {
+ *     await doSomethingTimeConsuming();
+ *     tryCancel();
+ *     await doSomethingElse();
+ *     tryCancel();
+ * }
+ * ```
+ */
+export function tryCancel(): void {
+    const { signal } = storage.getStore()?.cancelTask ?? {};
```
mnmkng, line 47: Maybe it's just me not being used to new JS features, but it took me a few secs to read it.

```
const signal = storage.getStore()?.cancelTask?.signal; // maybe?
```

packages/timeout/src/index.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        throw errorMessage instanceof Error
+            ? errorMessage
+            : new TimeoutError(errorMessage);
+    };
+
+    await new Promise((resolve, reject) => {
+        storage.run(context, () => {
+            Promise.race([timeout(), wrap()]).then(resolve, reject);
```
mnmkng, line 106: We used to do timeouts with `Promise.race()` and ended up with memory leaks, because even though the timeout fired and the race resolved, the other promise kept existing and it often held a lot of data, e.g. full page HTML. Please make sure this will not happen here.

test/timeout.test.ts
```diff
@@ -0,0 +1,89 @@
+describe('timeout with abort controller', () => {
+    it('empty', async () => {
+        // empty test so jest won't fail in node < 15 due to no tests being executed
+    });
+
+    if (Number(process.version.match(/v(\d+)/)?.[1] ?? null) < 15) {
```
mnmkng, line 6: Again, just a small readability nitpick. Personally, I'd rather create an extra variable than do stuff like `)?.[1] ?? null)`. Old school is cool.

```js
const [majorVersion] = process.version.split('.');
if (Number(majorVersion) < 15) {
```

test/timeout.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    beforeEach(() => {
+        position = 0;
+    });
+
+    it('passes without timeouting', async () => {
+        const res = await addTimeoutToPromise(
+            () => handler(),
+            200,
+            'timed out',
+        );
```
mnmkng, line 48: It might be worth using [fake timers](https://jestjs.io/docs/timer-mocks) to ensure the tests aren't flaky under load. But it might not be possible in this complex scenario.

packages/timeout/src/index.ts
B4nan: That's what I got initially, but we have linter that is forcing destructing assignments :D

packages/timeout/src/index.ts
B4nan: I dont think it will happen as we are explicitly aborting them (also I can see issues with the eager promise execution which is not present here). I did test this locally directly with SDK and puppeteer crawler. Will do more extensive testing once its handled in SDK master.

test/timeout.test.ts
B4nan: You still need to strip the `v` prefix, but sure, we can have some temp var to make it more readable. But as its inside tests I did not care much.

packages/timeout/src/index.ts
B4nan: Probably yes 😁

packages/timeout/src/index.ts
szmarczak: In that case I'd just `// eslint-disable-next-line ...`

test/timeout.test.ts
B4nan: Lets deal with that when it starts happening, it feels like it will be safer to test it works with real timeouts properly.

packages/timeout/src/index.ts
szmarczak: > We used to do timeouts with Promise.race() and ended up with memory leaks, because even though the timeout fired and the race resolved, the other promise kept existing and it often held a lot of data

This promise has no right to exist and should be collected by garbage collector. AFAIK, hanging promises are generally safe in JS. If it leaks then probably it's a bug in V8.

test/timeout.test.ts
szmarczak: `process.versions.node`

test/timeout.test.ts
szmarczak: I agree with @B4nan. We're using fake timers in tests in Got and some internal functions need to be patched. I spent hours trying to get them to work and in the end gave up. We can always run hundreds of small, real timeouts instead.

test/timeout.test.ts
mnmkng: > But as its inside tests I did not care much.

Ah, right, missed that while skimming the code.

> `process.versions.node`

This is what I meant 👍

test/timeout.test.ts
mnmkng: Well, we were not using fake timers in the SDK and we had A LOT of flaky tests 😄 But, ok, let's wait what happens 👍

packages/timeout/src/index.ts
mnmkng: See this issue https://github.com/apify/apify-js/issues/263 Not sure if it's still relevant, but at that time, it was an absolutely critical fix.

PR #268, opened 2021-11-23
fix!: rewrite the `addTimeoutToPromise` helper

B4nan:
During testing it turns out that in some places the previous implementation
did not work correctly, probably due to some bugs in ALS. To get around those,
we now use the same way as before with a regular `setTimeout/clearTimeout` for
the timeout handler. We no longer have `cancelTimeout` abort controller instance,
as we clear the timeout instead of using abort signal.

### DIFF
--- a/packages/timeout/src/index.ts
+++ b/packages/timeout/src/index.ts
@@ -89,7 +89,7 @@ export async function addTimeoutToPromise<T>(handler: () => Promise<T>, timeoutM
             reject(error);
         }, timeoutMillis);
 
-        exports.storage.run(context, () => {
+        storage.run(context, () => {
             wrap()
                 .then(() => resolve(returnValue))
                 .catch(reject)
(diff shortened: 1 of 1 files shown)


packages/timeout/src/index.ts
```diff
@@ -82,30 +79,21 @@ export async function addTimeoutToPromise<T>(handler: () => Promise<T>, timeoutM
...
 
-        throw errorMessage instanceof Error
-            ? errorMessage
-            : new TimeoutError(errorMessage);
-    };
-
-    await new Promise((resolve, reject) => {
-        storage.run(context, () => {
-            Promise.race([timeout(), wrap()]).then(resolve, reject);
+        exports.storage.run(context, () => {
```
vladfrangu, line 92:
```suggestion
        storage.run(context, () => {
```

vladfrangu (commented):
Small nit

packages/timeout/src/index.ts
B4nan: good catch, i was copy pasting from the compiled files from the repro and missed this one :D

PR #270, opened 2021-12-06
feat(hubspot_client): added few custom attributes

valekjo:


valekjo:
Closing, this won't be needed in the end.

PR #272, opened 2021-12-23
fix: consecutive dots in email are not allowed

katacek:
Solving LogDNA error bug, more info in Slack [here](https://apifier.slack.com/archives/G0134LE0JUB/p1638267910154500). 

In general:
Email with consecutive dots in it should not be allowed to register. It was tested by [Mailgun API](https://documentation.mailgun.com/en/latest/api-email-validation.html#request-examples) where those emails are marked as invalid.

The new regex was tested in MongoDB using the query below, now it returns 19 invalid emails, the same number as currently used regex:
`db.getCollection('users').find({
  $and: [
      { 'emails.address': { $not: { $regex: /NEW_REGEX/  } } },
      { 'emails.address': { $ne: null } },
  ]
})`

Moreover, new `invalid email` was added to tests.

Any comments / reviews appreciated :-) 
Thanks!

### DIFF
--- a/packages/consts/src/regexs.ts
+++ b/packages/consts/src/regexs.ts
@@ -1,8 +1,11 @@
 /**
  * Email validation regexp adapted from https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
  * with our restriction that hostname must be a TLD! (will not match example@localhost)
+ * and two consecutive dots in name are not allowed (based on Mailgun convention, will not match ex..amle@example.com)
  */
-export const EMAIL_REGEX_STR = '[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+(?:\\.[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+)*\\.{0,1}@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+'; // eslint-disable-line max-len
+const subRegexOne = 'a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-';
+const subRegexTwo = 'a-zA-Z0-9';
+export const EMAIL_REGEX_STR = `[${subRegexOne}]+(?:\\.[${subRegexOne}]+)*\\.{0,1}@[${subRegexTwo}](?:[${subRegexTwo}-]{0,61}[${subRegexTwo}])?(?:\\.[${subRegexTwo}](?:[${subRegexTwo}-]{0,61}[${subRegexTwo}])?)+`; // eslint-disable-line max-len
 
 /**
  * Matches a string containing valid email
--- a/packages/consts/src/regexs.ts
+++ b/packages/consts/src/regexs.ts
@@ -1,9 +1,3 @@
-/**
- * Email validation regexp adapted from https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
- * with our restriction that hostname must be a TLD! (will not match example@localhost)
- * and two consecutive dots in name are not allowed (based on Mailgun convention, will not match ex..amle@example.com)
- */
-
 // Parts for building an email regex (email will be constructed as `name@domain`)
 // name parts can be alnum + some special characters
 const namePartSubRegexStr = '[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+';
@@ -14,9 +8,18 @@ const domainPartSubRegexStr = '[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?'; /
 // doman is 2+ domain parts joined by periods (no leading or dangling period, no consecutive periods)
 const domainSubRegexStr = `${domainPartSubRegexStr}(?:\\.${domainPartSubRegexStr})+`;
 
-export const EMAIL_REGEX_STR = `^${nameSubRegexStr}@${domainSubRegexStr}$`;
+/**
+ * Email validation regexp adapted from https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
+ * with our restriction that hostname must be a TLD! (will not match example@localhost)
+ * and two consecutive dots in name are not allowed (based on Mailgun convention, will not match ex..amle@example.com)
+ */
+export const EMAIL_REGEX_STR = `${nameSubRegexStr}@${domainSubRegexStr}`;
 
-export const EMAIL_REGEX = new RegExp(EMAIL_REGEX_STR);
+/**
+ * Matches a string containing valid email
+ * Hostname must be a TLD! (will not match example@localhost)
+ */
+export const EMAIL_REGEX = new RegExp(`^${EMAIL_REGEX_STR}$`);
 
 /**
  * Comes from https://github.com/jonschlinkert/is-git-url/ but we have:
--- a/packages/consts/src/regexs.ts
+++ b/packages/consts/src/regexs.ts
@@ -2,7 +2,7 @@
  * Email validation regexp adapted from https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
  * with our restriction that hostname must be a TLD! (will not match example@localhost)
  */
-export const EMAIL_REGEX_STR = '[a-zA-Z0-9.!#$%&\'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+'; // eslint-disable-line max-len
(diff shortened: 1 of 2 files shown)


mtrunkat (changes requested):
I did a short test and it seems working well:

```js
var EMAIL_REGEX_STR = '[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+(?:\\.[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+)*\\.{0,1}@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+';
var OLD_EMAIL_REGEX_STR = '[a-zA-Z0-9.!#$%&\'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+';

db.getCollection('users').find({
  emails: {
      $ne: null,
      $not: { $size: 0 },
  },
  'emails.address': {
      $ne: null,
      $regex: new RegExp(`^${OLD_EMAIL_REGEX_STR}$`),
      $not: { $regex: new RegExp(`^${EMAIL_REGEX_STR}$`) }
  }     
}).map(u => u.emails[0].address)
```

obrazek

But I have a few comments:
- Please update a comment above regex describing our changes to the original regex
- Looking at tests I'd add a few more test cases such as 2 dots in a name `a.b.c+123~@example.com`
- There are some duplicate parts of this regex such as for example `[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+` and I think that it would make sense to extract them into separate sub-regex and test them separately
- Adding @valekjo and @fnesveda as they are better than me in this so perhaps they will have an idea how to simplify it

valekjo (commented):
I was a bit confused about the `(?:...)` but then I finally found what it is :D. But we probably need to keep those non capturing groups, if the regex str is used to match whole email.

I like @mtrunkat's idea about extracting parts of regex, but I'd do that just for the  "character classes" and give them a short name and long comment, so the regex would be more readable. eg: something like
```js
const X = `a-zA-Z0-9!#$%&'*+/=?^_\`{|}~-`;
const A = `a-zA-Z0-9`;
const EMAIL = `[${X}]+(?:\\.[${X}]+)*\\.{0,1}@[${A}](?:[${A}-]{0,61}[${A}])?(?:\\.[${A}](?:[${A}-]{0,61}[${A}])?)+`;
```
But now looking at it,not sure if it looks more readable or not.

mtrunkat (approved):
Good to me, but I'd wait for @fnesveda as he is our RegExp expert ;)

packages/consts/src/regexs.ts
```diff
@@ -1,8 +1,11 @@
 /**
  * Email validation regexp adapted from https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
  * with our restriction that hostname must be a TLD! (will not match example@localhost)
+ * and two consecutive dots in name are not allowed (based on Mailgun convention, will not match ex..amle@example.com)
  */
-export const EMAIL_REGEX_STR = '[a-zA-Z0-9.!#$%&\'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+'; // eslint-disable-line max-len
+const subRegexOne = 'a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-';
+const subRegexTwo = 'a-zA-Z0-9';
+export const EMAIL_REGEX_STR = `[${subRegexOne}]+(?:\\.[${subRegexOne}]+)*\\.{0,1}@[${subRegexTwo}](?:[${subRegexTwo}-]{0,61}[${subRegexTwo}])?(?:\\.[${subRegexTwo}](?:[${subRegexTwo}-]{0,61}[${subRegexTwo}])?)+`; // eslint-disable-line max-len
```
B4nan, line 8: `subRegexOne` is always used inside `[${subRegexOne}]+` so the variable could hold such value directly, and it could also leverage the `subRegexTwo` as it begins with the exact same sequence

also would be worth finding better names for the variables, or at least comment them

lastly, what about using just `\w` instead of `subRegexTwo`? it will also allow `_` in the group, which might be actually what we want (it is added to `subRegexOne` anyway, and I don't see any issues with having it part of `subRegexTwo` too

fnesveda (changes requested):
I also think it would be better to split the regex into multiple parts and name them and explain them better, how about something like this?

```javascript
// Parts for building an email regex (email will be constructed as `name@domain`)
// name parts can be alnum + some special characters
const namePartSubRegexStr = '[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+';
// name is 1+ name parts joined by periods (no leading or dangling period, no consecutive periods)
const nameSubRegexStr = `${namePartSubRegexStr}(?:\\.${namePartSubRegexStr})*`;
// domain parts can be alnum and dash characters (no leading and dangling dashes, max 63 chars long)
const domainPartSubRegexStr = '[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?'; //
// doman is 2+ domain parts joined by periods (no leading or dangling period, no consecutive periods)
const domainSubRegexStr = `${domainPartSubRegexStr}(?:\\.${domainPartSubRegexStr})+`;

export const EMAIL_REGEX_STR = `^${nameSubRegexStr}@${domainSubRegexStr}$`;

export const EMAIL_REGEX = new RegExp(EMAIL_REGEX_STR);
```

There's a few changes compared to your version:
- it does not allow trailing dots in the name (which is disallowed by the spec)
- it surrounds the whole regex by `^` and `$`, so it disallows partial matches to pass (e.g. `...test@example.com` would not pass now, but it would pass before)

We shoud probably add even more tests for the leading and trailing periods, and dashes in the domain name, and multiple subdomains.

packages/consts/src/regexs.ts
B4nan: Really? Never heard of that limitation. Good people of stack overflow says it is fine:

https://stackoverflow.com/questions/2180465/can-domain-name-subdomains-have-an-underscore-in-it

> Most answers given here are false. It is perfectly legal to have an underscore in a domain name. Let me quote the standard, RFC 2181, section 11, "Name syntax":
> 
> The DNS itself places only one restriction on the particular labels that can be used to identify resource records. That one restriction relates to the length of the label and the full name. [...] Implementations of the DNS protocols must not place any restrictions on the labels that can be used. In particular, DNS servers must not refuse to serve a zone because it contains labels that might not be acceptable to some DNS client programs.

katacek: Uff, not easy.. I went for the conservative approach and chose the @fnesveda regex proposal.

test/regexs.test.ts
```diff
@@ -119,13 +119,21 @@ const tests = {
...
+            'test@example.com ',
             'test@@example.com',
             'test@localhost',
             'not an email',
             '@example.com',
+            'test..test@example.com',
+            'test.@example.com',
+            '.test@example.com',
+            '...test@example.com',
+            'test@example_example.com',
```
B4nan, line 136: from what I understood, this is a valid domain, the only place where `_` is invalid is the top level domain (so the right most part after last dot), right @fnesveda? if so would be good to at least add a comment to this one in case we decide to support this at some point.

B4nan (approved):
looking good, only suggestion i can think of is maybe using case insensitive search so we can reduce `a-zA-Z` groups to just `a-z`

test/regexs.test.ts
fnesveda: Uff, I have no idea anymore, the info is conflicting. Let's add a comment for now and then maybe add support in the future in a separate PR, this one is becoming too long.

packages/consts/src/regexs.ts
```diff
@@ -1,14 +1,22 @@
...
+// name is 1+ name parts joined by periods (no leading or dangling period, no consecutive periods)
+const nameSubRegexStr = `${namePartSubRegexStr}(?:\\.${namePartSubRegexStr})*`;
+// domain parts can be alnum and dash characters (no leading and dangling dashes, max 63 chars long)
+const domainPartSubRegexStr = '[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?'; //
+// doman is 2+ domain parts joined by periods (no leading or dangling period, no consecutive periods)
+const domainSubRegexStr = `${domainPartSubRegexStr}(?:\\.${domainPartSubRegexStr})+`;
+
+export const EMAIL_REGEX_STR = `^${nameSubRegexStr}@${domainSubRegexStr}$`;
+
+export const EMAIL_REGEX = new RegExp(EMAIL_REGEX_STR);
```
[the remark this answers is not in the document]
fnesveda: Also, can you add back the JSDocs just before the `export const XXX` lines, so that they're associated with the constants, and show in editor hints etc.?

test/regexs.test.ts
katacek: comment `no underscore in domain name now but the standards are not 100% clear, might be reviewed` added

packages/consts/src/regexs.ts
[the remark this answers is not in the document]
katacek: thanks, done

fnesveda (approved):
Looking good! :)

valekjo (approved):

PR #281, opened 2022-02-02
feat(hubspot_client): add updateDeal method

novotnyj:
This will be used by MP to sync data from kanban with deal in hubspot.

I didn't add any checks/transformation to data. As it is not clear now and it will most likely change. I would rather do the changes in app... Later when it settles I will add `updateDealFromMarketplaceProject` or something and delete `updateDeal`...

closes [apify-core/#5130](https://github.com/apify/apify-core/issues/5130)

### DIFF
--- a/packages/hubspot_client/src/hubspot_client.ts
+++ b/packages/hubspot_client/src/hubspot_client.ts
@@ -769,6 +769,24 @@ export class HubspotClient {
         }
     }
 
+    /**
+     * Updates existing deal with given data
+     *
+     * @param {String|number} hubspotDealId
+     * @param {Object} data
+     */
+    async updateDeal(hubspotDealId: string | number, data: any): Promise<void> {
+        try {
+            await this.client.crm.deals.basicApi.update(`${hubspotDealId}`, {
+                properties: data,
+            });
+        } catch (_error) {
+            const error = _error as HttpError;
+            if (error.statusCode && error.statusCode === 404) throw new Error('Hubspot record not found');
+            throw error;
+        }
+    }
+
     /*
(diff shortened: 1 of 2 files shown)


gippy (approved):

packages/hubspot_client/src/hubspot_client.ts
```diff
@@ -769,6 +769,24 @@ export class HubspotClient {
         }
     }
 
+    /**
+     * Updates existing deal with given data
+     *
+     * @param {String|number} hubspotDealId
+     * @param {Object} data
```
B4nan, line 776: this is not adding any value with proper TS types, let's remove it
```suggestion
```

packages/hubspot_client/src/hubspot_client.ts
```diff
@@ -769,6 +769,24 @@ export class HubspotClient {
...
+     */
+    async updateDeal(hubspotDealId: string | number, data: any): Promise<void> {
```
B4nan, line 778: if `data` can be any object, we can use `Record<string, any>`, or even better `Record<string, unknown>`, given we are not working with the properties so we dont care about their type
```suggestion
    async updateDeal(hubspotDealId: string | number, data: Record<string, unknown>): Promise<void> {
```

novotnyj: Thanks!

B4nan: looks like it broke the tests because the usage expects strings values, so lets change it to `Record<string, string>`

B4nan (approved):

PR #288, opened 2022-03-14
feat: add `pseudo_url` package

AndreyBykov:
I will follow up with `apify-ts` update once I merge the updates there (I see there were changes to `enqueue_links` a few days ago) + finish with tests. But otherwise, I think if this one's good - I don't really need anything else here - in SDK it will only be `new PseudoUrl()` and I will just get `regex` from it and would not even use `matches` from `apify-shared-js`, thus `createRequests` will use plain `match` as pseudo urls and globs will be converted to regexps (+ direct regexps option).
As discussed - no request related things in this package compared to current implementation in `apify-js` repo.


AndreyBykov:
Current implementation in `apify-js` for reference: https://github.com/apify/apify-js/blob/master/src/pseudo_url.js

packages/pseudo_url/src/pseudo_url.ts
```diff
@@ -0,0 +1,127 @@
...
+    } catch (err) {
+        throw new Error(`Cannot parse PURL '${purl}': ${err}`);
+    }
+
+    return new RegExp(regex, 'i');
+}
+
+export interface PseudoUrlObject {
+    purl: string | RegExp;
+}
```
B4nan, line 127: seems to be unused

B4nan: it should be in the SDK if its used there, not here

AndreyBykov:
I should also add/move the tests, will do it first thing in the morning... I somehow kept it in mind for sdk repo :/

B4nan:
The lock file seems to be wrong, it contains whole SDK.

B4nan (approved):

PR #293, opened 2022-04-27
feat(hubspot_client): add support for kanban url to contact

novotnyj:
We want to set `Apify kanban link` property at HS contact to the link to the latest MP project.

### DIFF
--- a/packages/hubspot_client/src/hubspot_client.ts
+++ b/packages/hubspot_client/src/hubspot_client.ts
@@ -38,6 +38,7 @@ const ALLOWED_USER_FIELDS = {
     billing_email__c: _.isString,
 
     lifecyclestage: _.isString,
+    apify_kanban_link_: _.isString,
 };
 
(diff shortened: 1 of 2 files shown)


packages/hubspot_client/src/hubspot_client.ts
```diff
@@ -38,6 +38,7 @@ const ALLOWED_USER_FIELDS = {
     billing_email__c: _.isString,
 
     lifecyclestage: _.isString,
+    apify_kanban_link_: _.isString,
```
B4nan, line 41: is the trailing underscore intentional?

novotnyj: Unfortunately yes 😄  The field is named `Apify kanban link:` in Hubspot - they seem to replace non-alphanum chars with underscores...

gippy (approved):

PR #303, opened 2022-05-25
feat: remove @apify/hubspot_client package

valekjo:
We're moving hubspot client to @apify-packages/hubspot-client to make things more simple.

The respective PR is here: https://github.com/apify/apify-core/pull/6549

@B4nan - is this ok for the package removal? Was not sure if I should use `feat` or `chore` or something else.
Also, I am not sure about the changes in lockfile, could you take a quick look on that (added `dev: true`s)


B4nan (approved):
> is this ok for the package removal? Was not sure if I should use feat or chore or something else.

Doesn't really matter, you remove the package, but it stays on NPM. And no packages are updated so there shouldn't be any release from this commit (maybe there will be, but not intentionally, the script does not work 100% correct iirc).

> Also, I am not sure about the changes in lockfile, could you take a quick look on that (added dev: trues)

idk, you are not the first one with this problem, i guess it does not matter much, the lock file will be regenerated on next release anyway

PR #305, opened 2022-05-31
fix(consts): remove g flag from regex

novotnyj:
For some reason my webstrom adds a `g` flag by default and I didn't noticed 😎


jancurn:
@novotnyj Please can we add a unit test to cover this error, to avoid any regressions?

PR #309, opened 2022-06-03
feat(consts): add test for 'g' flag in COMMA_SEPARATED_EMAILS_REGEX

novotnyj:
Adds test sugested here https://github.com/apify/apify-shared-js/pull/305#issuecomment-1145824773


test/regexs.test.ts
```diff
@@ -301,6 +301,11 @@ describe('regexps', () => {
             });
         });
     });
+    // The 'g' flag causes RegExp object to keep track of last match occurrence in `lastIndex`.
```
jancurn, line 304: I meant to test the effect of the `g` flag that prevented the correct usage of the regexp in the failed deploy, rather than just testing the definition :)

PR #312, opened 2022-06-16
feat: native ESM all the packages

vladfrangu:
While all the compiled code looks fine and functional, testing would still be needed to ensure it still works correctly


vladfrangu:
Ouch... Shapeshift is node 16+ only it seems... I thought nodejs 14 also had it. Might have to either experiment with ow or make shapeshift compatible with node 14

B4nan:
Personally i dont mind node 16 either, those are internal packages so as long as we run things on node 16 internally… but i am not sure about that tbh, i would expect older version

vladfrangu:
I think a huge chunk internally is still on node 14, so yeah.. will have to look into this

vladfrangu:
Actually, I don't get why the CI is failing on node 14, as it supports `?.` and `??`...

vladfrangu:
We made shapeshift be nodejs 14+ compatible, and I feel like we could drop the nodejs 12 CI here, especially since it's no longer even supported by node. Thoughts?

mnmkng:
@fnesveda @gippy Would be great if you guys could chime in. I'm not sure if this won't break anything on the platform :)

vladfrangu:
If platform depends in any way on the ow errors, it will break *something*. Otherwise it should just work the same, but I do wonder if it can be tested somehow 👀

fnesveda:
@vladfrangu @mnmkng So what's actually the goal of this PR? Would be nice to get some short description of what has changed instead of just a review request out of the blue, so we don't have to play detective in 43 changed files.

mnmkng:
@fnesveda It's a change to use native ESM imports and modules instead of CJS AFAIK.

I added you because I had no idea if Platform even supports ESM.

fnesveda (approved):
If it still builds CJS versions as well, and they are the default under `main` in `package.json`, then it should be importable / requireable on the platform.

We'll just have to change a few imports since all the files are now bundled, and we can't import from them directly, but nothing major.

Could you release this as a major version, just to make sure we don't accidentally update the packages with Renovate bot etc.?

fnesveda:
Also CC @jbartadev we'll probably have to update web to Node 14+ because of this (and other reasons).

B4nan (approved):

.github/workflows/test_and_release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                node-version: [ 12, 14, 15, 16 ]
+                node-version: [ 14, 15, 16, 18 ]
```
B4nan, line 18:
```suggestion
                node-version: [ 14, 16, 18 ]
```

B4nan:
Let's see how it works... 🤞

B4nan:
Looks like it works fine, already added to crawlee, E2E tests are passing including on platform.

PR #316, opened 2022-06-30
feat(log): provide a `warningOnce` alias for `log.deprecated`

mvolfik:


mnmkng (approved):

packages/log/src/log.ts
```diff
@@ -255,12 +255,19 @@ export class Log {
...
+    warningOnce(message: string) {
+        if (this.warningsOnceLogged.has(message)) return;
 
-        this.deprecationsReported[message] = true;
+        this.warningsOnceLogged.add(message);
         this.warning(message);
     }
+
+    /**
+     * Alias for log.warningOnce() kept for backwards compatibility
```
B4nan, line 268: lets keep the comment as it was, this method is not just alias for BC, its valid public API 

```suggestion
     * Logs given message only once as WARNING. It's used to warn user that some feature he is using has been deprecated.
```

PR #319, opened 2022-07-18
feat(markdown): add GraphQL

mstephen19:


mstephen19:
Also added `quote_type` to `.editorconfig`

packages/markdown/src/marked.ts
```diff
@@ -150,9 +162,11 @@ export const apifyMarked = (markdown: string): MarkedResponse => {
     tokens.forEach((token) => {
         if (token.type === 'code' && token.lang) {
             if (token.lang === 'marked-tabs') {
-                codeTabsObjectPerIndex[markedTabTokenIndex] = codeTabObjectFromCodeTabMarkdown(token.text);
+                codeTabsObjectPerIndex[markedTabTokenIndex] =
```
B4nan, line 165: like this line, why are you breaking it? my guess is prettier and their dumb 80 chars max line length default? :]

B4nan (commented):
Would be nice to separate code style changes into separate PR, as its hard to see what you actually changed and what is just code formatting. Also there are some trailing commas missing and lint step fails.

TBH those line breaks look like from prettier. And prettier does quite a poor job in adding line breaks if you ask me :]

packages/markdown/src/marked.ts
mstephen19: @B4nan yeah sorry that's prettier. I have auto-save when a window is unfocused

mstephen19: But I have the line length limit to 140

B4nan: Using any auto-formatting on projects that does not provide configuration for the tool is... Let's say not very polite :] Even if we want to change code style, it can't be part of such a PR (that should have pretty much one line diff), it needs to be a separate one. 

> But I have the line length limit to 140

That line had 107 characters 🤷

fnesveda:
I was going through old, forgotten PRs to clean them up, and I decided to finish this one, since it's tiny but useful.

PR #334, opened 2022-09-14
feat(utilities): add publicEncrypt() and privateDecrypt() method

drobnikj:
New methods, which we will use for secure input and for secure env vars in the future.
see [this slack message](https://apifier.slack.com/archives/C041VJB1WSC/p1663145065980509) for some notes 

Notes: 
* I use [aes-gcm over aes-ctr](https://crypto.stackexchange.com/questions/44764/relationship-between-aes-gcm-and-aes-ctr) because in GCM you can use authenticated encryption
* [Authenticated encryption](https://en.wikipedia.org/wiki/Authenticated_encryption) is good to integrity check of cipher to away from attack with prepend or append bytes to your cipher
* I generate IV for each encrypted value to avoid [nonce disrespect](https://github.com/nonce-disrespect/nonce-disrespect)

I have attached some links, I hope it helps to understand, I struggled it in a little for a while until I found it.


packages/utilities/src/crypto.ts
```diff
@@ -0,0 +1,68 @@
+import crypto from 'crypto';
```
fnesveda, line 1: We are importing the `@apify/utilities` package on the frontend, so adding `crypto` as a dependency will cause the import to break, or we will require a huge polyfill. I would put these utilities in a separate package.

packages/utilities/src/crypto.ts
```diff
@@ -0,0 +1,68 @@
...
+ * and uses the password(consists of encrypted key and initial vector)
+ * to decrypt the encrypted value.
+ *
+ * @param privateKey {Buffer} Private key used for decryption
+ * @param passphrase {string} Passphrase from private key
+ * @param encryptedPassword {string} Password in Base64 encrypted using private key
+ * @param encryptedValue {string} Content in Base64 encrypted using AES cipher
+ * @returns {string}
+ */
+export function privateDecrypt(privateKey: Buffer, passphrase: string, encryptedPassword: string, encryptedValue: string): string {
```
fnesveda, line 49: Let's put these parameters as an object, otherwise someone will surely write them in the wrong order, and it will be a pain in the ass to debug, as these cryptographic things usually are.

packages/utilities/src/crypto.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * It encrypts the given value using AES cipher and the password for encryption using the public key.
+ * NOTE: The encryption password is a string of encryption key and initial vector used for cipher.
+ * It returns the encrypted password and encrypted value in BASE64 format.
+ *
+ * @param publicKey {Buffer} Public key used for encryption
+ * @param value {string} Value to be encrypted
+ * @returns {Object<encryptedPassword, encryptedValue>}
+ */
+export function publicEncrypt(publicKey: Buffer, value: string) {
```
fnesveda, line 18: Same here, even though there's just two of them.

packages/utilities/src/crypto.ts
```diff
@@ -0,0 +1,68 @@
...
+        throw new Error('privateDecrypt: Decryption failed, invalid password length!');
+    }
+
+    // Slice Auth tag from the final value cipher
+    const authTagBuffer = encryptedValueBuffer.slice(encryptedValueBuffer.length - ENCRYPTION_AUTH_TAG_LENGTH);
+    const encryptedDataBuffer = encryptedValueBuffer.slice(0, encryptedValueBuffer.length - ENCRYPTION_AUTH_TAG_LENGTH);
+
+    const encryptedKeyBuffer = passwordBuffer.slice(0, ENCRYPTION_KEY_LENGTH);
+    const initVectorBuffer = passwordBuffer.slice(ENCRYPTION_KEY_LENGTH);
+    const decipher = crypto.createDecipheriv(ENCRYPTION_ALGORITHM, encryptedKeyBuffer, initVectorBuffer);
```
fnesveda, line 64: The key is not encrypted at this point anymore, right?
```suggestion
    const encryptionKeyBuffer = passwordBuffer.slice(0, ENCRYPTION_KEY_LENGTH);
    const initVectorBuffer = passwordBuffer.slice(ENCRYPTION_KEY_LENGTH);
    const decipher = crypto.createDecipheriv(ENCRYPTION_ALGORITHM, encryptionKeyBuffer, initVectorBuffer);
```

test/crypto.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    it('should return different cipher for the same string', () => {
+        const randomString = utils.cryptoRandomObjectId(100);
+        const uniqueCiphers = new Set();
+        const uniqueCiphers16Bytes = new Set();
+        for (let i = 0; i < 100; i++) {
+            const { encryptedPassword, encryptedValue } = utils.publicEncrypt(publicKey, randomString);
+            const theFirst16Bytes = Buffer.from(encryptedValue, 'base64').toString('utf-8').slice(0, 16);
+            expect(uniqueCiphers.has(encryptedPassword)).toBe(false);
+            expect(uniqueCiphers16Bytes.has(theFirst16Bytes)).toBe(false);
+            uniqueCiphers.add(encryptedValue);
```
fnesveda, line 53: Are you checking the uniqueness of `encryptedValue`, or `encryptedPassword`?

fnesveda (changes requested):
Overall looks good, just a few small notes 🙂

packages/utilities/src/crypto.ts
fnesveda: Hm, OK, maybe it's not necessary to move then, tree shaking probably takes care of it. Let's keep it like this 🙂

test/crypto.test.ts
drobnikj: encryptedValue, there is a typo, will fix it.

fnesveda (approved):

jirimoravcik (approved):

PR #335, opened 2022-09-26
feat(input_schema): add isSecret flag into the sting property

drobnikj:
Add support for secure input values by adding `isSecret` flag into the sting property

The flag is optional, so the change is not breaking.

EDIT:

I improved validation, if isSecret is `true` fields:
`['minLength', 'maxLength','default', 'prefill', 'pattern']`
are not in string type definition and editor is only enum for 
`['textfield', 'textarea', 'hidden']`


mtrunkat (approved):
How does it going to work with other options such as:
- `editor`? We could display the correct editor later on in the modal? This is useful, for example, for some secret JSON values.
- we should probably disable max/min length as it could collide with encrypted data or no?

fnesveda (approved):

drobnikj:
@mtrunkat I cannot find any way how to do conditional fields in json schema. As you said, it would be good, but not sure how to do it in schema.

fnesveda:
@drobnikj JSON schema now supports if-then-else (since draft-07), maybe that would work?
https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else

fnesveda:
Also, if the conditional fields in the schema are possible, maybe we should disable `prefill` and `default` for secret fields too, it would make the frontend code a lot easier, and I don't think it really hurts any usecase.

drobnikj:
I have quite hard times finding way how to do it. It is a little bit breakneck, but it should work.

drobnikj:
I updated the validation, please check again @fnesveda @mtrunkat

packages/input_schema/src/schema.json
```diff
@@ -79,23 +79,61 @@
...
-                "prefill": { "type": "string" },
                 "example": { "type": "string" },
-                "pattern": { "type": "string" },
                 "nullable": { "type": "boolean" },
-                "minLength": { "type": "integer" },
-                "maxLength": { "type": "integer" },
                 "editor": { "enum": ["javascript", "python", "textfield", "textarea", "hidden"] },
+                "isSecret": { "type": "boolean" },
                 "sectionCaption": { "type": "string" },
                 "sectionDescription": { "type": "string" }
```
fnesveda, line 92: Maybe you can delete everything but `isSecret`? It will get overriden anyway in the `then` and `else` directives, and if you remove it, it will make the returned errors a bit more readable.

fnesveda (approved):
Uff, nice job!

I thought it would be easier, but the `additionalProperties` rule is making it complicated.

packages/input_schema/src/schema.json
drobnikj: I remove some of them, but I need to keep these because we used them in some of our custom validation methods or in error generation and for checking required fields.

fnesveda: Ah, right, I was checking it just with AJV, not with our custom validations.

jirimoravcik (approved):

PR #340, opened 2022-10-05
feat(input_secrets): add new package to encrypt and decrypt input secrets

drobnikj:
Move the code from https://github.com/apify/apify-core/pull/8266

We need to use `decryptInputSecrets` and in the future `encryptInputSecrets` in apify-sdk, so I move it here.

### DIFF
--- a/packages/input_secrets/package.json
+++ b/packages/input_secrets/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@apify/input_secrets",
     "version": "1.0.0",
-    "description": "Tool to find the matching URLs on a page (or html document) by providing a URL pattern.",
+    "description": "Package to encrypt and decrypt apify actor input secrets.",
     "main": "./dist/index.js",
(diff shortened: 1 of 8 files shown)


fnesveda (approved):

packages/input_secrets/package.json
```diff
@@ -0,0 +1,46 @@
+{
+    "name": "@apify/input_secrets",
```
jirimoravcik, line 2: why underscore in the name?

jirimoravcik (approved):

packages/input_secrets/package.json
drobnikj: It needs to be same as folder path because https://github.com/apify/apify-shared-js/blob/master/tsconfig.json#L6

jirimoravcik: ok, understandable :D still it's common to have a dash instead of underscore

PR #346, opened 2022-10-13
fix(input_schema): import from commonjs package

nguyeda1:
The same issue as https://github.com/apify/apify-shared-js/pull/345


nguyeda1:
Nevermind I will just transpile it to commonjs in our repo

PR #353, opened 2022-12-01
chore(ci): replace `npm ci` with npm `i` in publish step

valekjo:


vladfrangu (approved):

B4nan (changes requested):
Lets regenerate the lockfile too fix the install step.

valekjo:
@B4nan I've replaced it in all install steps, did you mean that?

vladfrangu:
> @B4nan I've replaced it in all install steps, did you mean that?

I don't think so, I think he meant run `npm i` locally and push the updated package-lock.json in this PR 👀

B4nan:
One more thing, run `npx lerna ls --json | npx ts-node -T scripts/sync-root-changelog.ts` to update the changelog in root, that's the last step that wasn't done because of the failure.

valekjo:
@B4nan done.

PR #369, opened 2023-03-23
feat: allow using variables in fields of webhook payload template

valekjo:
This PR
 - makes parsing slightly more effective (one pass only)
 - adds support for "interpolation" - using `{{var:resource.defaultDatasetId}}` syntax in strings - ie. adding the possibility to pass it to the input.
 - It's intentionally slightly different mechanism than the parsing, as it does a different thing.
 
 Theoretically it's breaking, I have an idea how to deal with it, just not sure if it's necessary.

### DIFF
--- a/packages/utilities/src/webhook_payload_template.ts
+++ b/packages/utilities/src/webhook_payload_template.ts
@@ -88,6 +88,7 @@ export class WebhookPayloadTemplate {
         if (type !== 'string') throw new Error(`Cannot parse a ${type} payload template.`);
         const template = new WebhookPayloadTemplate(payloadTemplate, allowedVariables, context);
         const data = template._parse(); // eslint-disable-line no-underscore-dangle
+        // TODO: Maybe make this configurable via some options, to make "sure" about backwards compatibility?
         return template._interpolate(data); // eslint-disable-line no-underscore-dangle
     }
 
@@ -176,7 +177,6 @@ export class WebhookPayloadTemplate {
         return value;
     }
 
-    // TODO: Just replace the variables in this case
     private _interpolateString(value: string): string {
         // If the string matches exactly, we return the variable value including the type
         if (value.match(/^\{\{var:([a-zA-Z0-9.]*)\}\}$/)) {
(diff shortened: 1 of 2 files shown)


packages/utilities/src/webhook_payload_template.ts
```diff
@@ -125,31 +133,74 @@ export class WebhookPayloadTemplate {
...
+        if (typeof value === 'object' && value !== null) {
+            return this._interpolateObject(value);
+        }
+        // We can't interpolate anything else
+        return value;
+    }
+
+    private _interpolateString(value: string): string {
+        // If the string matches exactly, we return the variable value including the type
+        if (value.match(/^\{\{var:([a-zA-Z0-9.]*)\}\}$/)) {
```
fnesveda, line 172: This would make `{{var:}}` valid too, you should have `+` instead of `*` in the regex

fnesveda (changes requested):
Looks good, just a few comments mostly about the tests. Plus we talked live about maybe dropping the `var:` prefix.

packages/utilities/src/webhook_payload_template.ts
valekjo: fixed

valekjo:
> Looks good, just a few comments mostly about the tests. Plus we talked live about maybe dropping the `var:` prefix.

@fnesveda I've updated the tests, found one bug during the process (how replacement works). And figured out one more reason why I added the `var:` prefix - so it doesn't break the original behaviour. So instead of `var:` there is new `interpolateStrings` option, which just turns on/off the parsing in strings.

packages/utilities/src/webhook_payload_template.ts
```diff
@@ -125,31 +140,75 @@ export class WebhookPayloadTemplate {
...
+            return this._interpolateObject(value);
+        }
+        // We can't interpolate anything else
+        return value;
+    }
+
+    private _interpolateString(value: string): string {
+        // If the string matches exactly, we return the variable value including the type
+        if (value.match(/^\{\{([a-zA-Z0-9.]+)\}\}$/)) {
+            // This just strpis the {{ and }}
```
fnesveda, line 180:
```suggestion
            // This just strips the {{ and }}
```

fnesveda (approved):
Looks good, one tiny typo

gippy (approved):

PR #376, opened 2023-04-11
feat!: Removing unused ACTOR_BASE_DOCKER_IMAGES and ACTOR_BASE_DOCKER_IMAGE_DEFAULT constants

mtrunkat:
This constant is not currently used anywhere, and the repo with docker images should serve as a single source of truth.


packages/consts/src/consts.ts
```diff
@@ -75,139 +75,9 @@ export const META_ORIGINS = {
...
-        displayName: '[DEPRECATED] BETA: Node.js 10 + Puppeteer on Debian - use apify/actor-node-chrome:beta instead!',
-        copyChown: 'node:node',
-    },
-];
-
-/**
- * Default image from ACTOR_BASE_DOCKER_IMAGES.
- */
-export const ACTOR_BASE_DOCKER_IMAGE_DEFAULT = ACTOR_BASE_DOCKER_IMAGES[0].name;
+export const ACTOR_BASE_DOCKER_IMAGE_DEFAULT = 'apify/actor-node';
```
fnesveda, line 80: Maybe we can remove this const too? We have no notion of "default base Docker image" anymore, and it's only used in two files in integration tests, where we can/should hardcode it anyway.

fnesveda (changes requested):
Looks good, but we probably can remove the other const too.

packages/consts/src/consts.ts
mtrunkat: Yeah, right, I see it's only used on tests and there it can be hardcoded for sure.

mtrunkat: Done, fixed in https://github.com/apify/apify-core/pull/10579

fnesveda (approved):

PR #383, opened 2023-05-12
fix(input_schema): proxygroup validation

katacek:
Fixing wrong proxygroup validation (if the group is array of strings), there is a[ connected PR for the message](https://github.com/apify/apify-core/pull/10953/files)

Working for our case, I have checked also for general input schema, if it returns correct values for different schema amd it seems ok.

Screenshot 2023-05-15 at 13 23 46

Screenshot 2023-05-15 at 13 41 51


packages/input_schema/src/utilities.ts
```diff
@@ -69,7 +69,10 @@ function validateProxyField(
     // If options are not provided skip additional checks
     if (!options) return fieldErrors;
 
-    const selectedProxyGroups = (apifyProxyGroups || []);
+    let selectedProxyGroups = [];
+    if (apifyProxyGroups) {
+        selectedProxyGroups = Array.isArray(apifyProxyGroups) ? apifyProxyGroups : [apifyProxyGroups];
```
fnesveda, line 74: I think it would be better to be strict here, we should check that if `apifyProxyGroups` is defined, it is an array of strings, and throw an error otherwise. This would still let through numbers, objects and other non-string things, which would then create a mess later.

fnesveda (changes requested):
Good idea to solve it so simply, but I think it won't work unfortunately 🙁

fnesveda (approved):
Nice!

PR #392, opened 2023-05-29
feat: allow custom proxies to be https

gippy:
Ondra needs to have custom proxy urls to allow https since he want's to test some provider proxies and they do not have http. He says it's working in crawlee, but our input schema does not allow it. I checked and it looks like this regex is only used for that one check.


B4nan (approved):
(expect the broken CI, but that's just about fixing the old test assertions)

PR #398, opened 2023-06-28
feat: add "ACTOR_" env vars

jirimoravcik:
Some stuff to discuss:
1. Is it better to create a separate dict `ACTOR_ENV_VARS` or should it just be placed inside `ENV_VARS` with some weird prefix? (since there are already variables like `ACTOR_RUN_ID` which points to `APIFY_ACTOR_RUN_ID`)
2. `ACTOR_MAX_PAID_DATASET_ITEMS` is in the original `ENV_VARS` dict and it's the only one not prefixed with `APIFY_`, which is quite inconsistent. Should it be moved/duplicated? (and it's also missing from the actor spec)


fnesveda (approved):
Yeah, the naming will always be a mess here, I'm afraid. I think having a second dict is the best choice, and I would move the `ACTOR_MAX_PAID_DATASET_ITEMS` there (cc @mhamas).

Not sure how we'll deal with this on the SDK side though, consuming and merging these env vars will be a pain.

mhamas:
sounds good to me, just please work closely with delivery and tooling team so that actors are not broken.

drobnikj (approved):

PR #399, opened 2023-07-04
fix: throw when `privateDecrypt` fails for input secrets

jirimoravcik:
Comments regarding the error text message are welcome.


packages/input_secrets/src/input_secrets.ts
```diff
@@ -69,7 +69,11 @@ export function decryptInputSecrets<T>(
             const match = value.match(ENCRYPTED_INPUT_VALUE_REGEXP);
             if (!match) continue;
             const [, encryptedPassword, encryptedValue] = match;
-            decryptedInput[key] = privateDecrypt({ privateKey, encryptedPassword, encryptedValue });
+            try {
+                decryptedInput[key] = privateDecrypt({ privateKey, encryptedPassword, encryptedValue });
+            } catch {
+                throw new Error(`The input field "${key}" could not be decrypted. You can fix this by updating the field's value in the input editor.`);
```
drobnikj, line 75: Please add simple test https://github.com/apify/apify-shared-js/blob/master/test/input_secrets.test.ts#L36

drobnikj (changes requested):
Is swallowing all errors a good idea? It can happen that changing input cannot help. 
What about at least log the original error? Otherwise, it is very hard to debug in case of any unexpected issue.

packages/input_secrets/src/input_secrets.ts
valekjo, line 75: To get around the swallowing of the original error, maybe we could add the original error as `cause` to the new one, so it would be accessible?

And for the message I'd go with something like this - not saying it fixes it, just that it might, and giving some more details.

`The input field "${key}" could not be decrypted. Try updating the field's value in the input editor. Decryption error: ${err}`

test/input_secrets.test.ts
```diff
@@ -48,4 +48,12 @@ describe('input secrets', () => {
         const encrypted2 = encryptInputSecrets({ input: encrypted1, inputSchema, publicKey });
         expect(testInput).toStrictEqual(decryptInputSecrets({ input: encrypted2, privateKey }));
     });
+
+    it('should throw if private key is not valid', () => {
+        const testInput = { secure: 'a secret string', customString: 'another string' };
+        const encryptedInput = encryptInputSecrets({ input: testInput, inputSchema, publicKey });
+        expect(encryptedInput.secure).not.toEqual(testInput.secure);
+        expect(encryptedInput.customString).toEqual(testInput.customString);
+        expect(() => decryptInputSecrets({ input: encryptedInput, privateKey: publicKey })).toThrow();
```
drobnikj, line 57: Can you check that the message is as expected?

drobnikj (changes requested):

valekjo (approved):
👍

drobnikj (approved):

PR #403, opened 2023-07-11
feat: add more ACTOR_ env vars, add a test, rename ENV_VARS

jirimoravcik:
I added more actor env vars based on updated https://github.com/apify/actor-specs/#environment-variables

I added a test that checks if all actor env vars begin with `ACTOR_`.

I renamed `ENV_VARS` to `APIFY_ENV_VARS` and created a new `ENV_VARS` variable that just copies the values for backwards compatibility. Also added a test to check if this works correctly.

### DIFF
--- a/test/consts.test.ts
+++ b/test/consts.test.ts
@@ -1,4 +1,4 @@
-import { USERNAME, APIFY_ID_REGEX } from '@apify/consts';
+import { USERNAME, APIFY_ID_REGEX, ACTOR_ENV_VARS, ENV_VARS, APIFY_ENV_VARS } from '@apify/consts';
 import { cryptoRandomObjectId } from '@apify/utilities';
 
 describe('consts', () => {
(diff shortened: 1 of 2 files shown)


test/consts.test.ts
```diff
@@ -49,4 +49,20 @@ describe('consts', () => {
             }
         });
     });
+
+    describe('ACTOR_ENV_VARS', () => {
+        it('every value begins with "ACTOR_"', () => {
+            Object.values(ACTOR_ENV_VARS).forEach((v) => {
+                expect(v.startsWith('ACTOR_')).toBe(true);
```
fnesveda, line 56: I think we can be stricter with the test
```suggestion
        it('every value is "ACTOR_" + key', () => {
            Object.entries(ACTOR_ENV_VARS).forEach(([k, v]) => {
                expect(v).toBe(`ACTOR_${k}`);
```

test/consts.test.ts
```diff
@@ -49,4 +49,20 @@ describe('consts', () => {
...
+        });
+    });
+
+    describe('APIFY_ENV_VARS', () => {
+        it('is the same as ENV_VARS', () => {
+            Object.keys(APIFY_ENV_VARS).forEach((k) => {
+                expect(APIFY_ENV_VARS[k]).toBe(ENV_VARS[k]);
+            });
+        });
+    });
```
fnesveda, line 67: Here we also can be stricter
```suggestion

        it('every value is "APIFY_" + key', () => {
            Object.entries(APIFY_ENV_VARS).forEach(([k, v]) => {
                // TODO: remove this once ACTOR_MAX_PAID_DATASET_ITEMS is removed from APIFY_ENV_VARS
                if (k === 'ACTOR_MAX_PAID_DATASET_ITEMS') return;

                expect(v).toBe(`APIFY_${k}`);
            });
        });
    });
```

fnesveda (changes requested):
Looks good, I just had an idea about making the tests stricter

test/consts.test.ts
jirimoravcik: smart, nice

fnesveda (approved):

drobnikj (approved):

PR #408, opened 2023-08-09
feat(utilities): implement signing and verification for the "run on apify" links

### LINKED ISSUE #400: Support signing/verifying in `encodeInput` and `decodeInput` (for run on apify button)
We need to be able to sign the input URL hash so we can verify the example comes from a trusted source, and its safe to run it right ahead.

B4nan:
Allows hashing of an Actor input together with some metadata into a shareable link for the "Run on Apify" button. Uses a common secret for checking the signatures.

The hash consists of 3 parts separated by a dot, as in `ABC.DEF.GHI`, each being a base64url encoded string:
 - `meta` object with the `version` and `user` properties.
 - `data` object (the one that gets encoded)
 - `signature` used for verification of the URL hash, computed from the `meta` and `data` objects

```ts
const manager = new CodeHashManager(secret);
const hash = manager.encode(input, '123'); // '123' is a user id

console.log(hash);

const { data, meta } = manager.decode(hash);

expect(input).toEqual(data);
expect(meta).toEqual({
    version: 1,
    user: '123',
    validSignature: true,
});
```

Closes #400


valekjo (approved):
Nice, thanks.

PR #413, opened 2023-09-13
feat(consts): add INTEGRATIONS actor category

valekjo:


fnesveda (approved):

HonzaKirchner (approved):

B4nan:
The build is apparently broken (probably some dependency update that requires newer node versions), without that the publishing won't work either... 

Ideally, I would just require node 18 everywhere and ship new major versions for every package, but I guess you still need node 16 supported for the console/api/etc?

valekjo:
> The build is apparently broken (probably some dependency update that requires newer node versions), without that the publishing won't work either...
> 
> Ideally, I would just require node 18 everywhere and ship new major versions for every package, but I guess you still need node 16 supported for the console/api/etc?

We still need 14 for meteor :D

B4nan:
We already discussed this with Pepa, IIRC we need npm@7 to have workspaces working, we don't need latest. Since it was working fine up until now, I would probably install v9 instead of v7.

valekjo:
v9 also works, will go with that one.

PR #428, opened 2024-01-02
chore: log cause when logging exceptions

### LINKED ISSUE #427: Improve logging of Error objects
Currently, when we log an error, the `cause` property is missing. We could use `util.inspect` in the logger package which should resolve it.

vladfrangu:
Closes #427 

Doing it this way to make it as non-breaking as possible

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "module": "commonjs",
-    "target": "es2020",
+    "target": "es2022",
     "moduleResolution": "node",
     "declaration": true,
     "sourceMap": false,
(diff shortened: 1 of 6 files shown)


tsconfig.build.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "module": "commonjs",
-    "target": "es2020",
+    "target": "es2022",
```
B4nan, line 4: isn't this breaking?

vladfrangu: We still build for es2020 (check tsup config), it's just the type level support for es2022 (since cause is an es2022 prop)

Alternative is...code checks and casts, which I can do, just let me know

B4nan: ok sounds fine

B4nan (approved):

PR #433, opened 2024-01-11
feat: make @apify/log browser-compatible

### LINKED ISSUE #432: Browser compatibility for `@apify/log` (and more?)
The `@apify-packages/ui-components` package (see [source here](https://github.com/apify/apify-core/tree/develop/src/packages/ui-components)) transitively depends on various packages from this repository, most notably `@apify/log`, `@apify/consts` and `@apify/utilities`. 

Because of this, we have to deal with annoying Webpack issues (see [thread here](https://github.com/apify/apify-docs/issues/809#issuecomment-1880922762)). 

An educated guess (see the thread for details) suggests it should be enough to make `@apify/log` a bit more browser-friendly - i.e. null checks on property access, check the `node:util` dependency etc.

barjin:
With these changes, `@apify/log` should run in a browser just fine... except for the `node:util` import, which has to be disabled with `webpack`'s `resolve.fallback.util: false`. Is there a better way of doing a conditional import (yeah, I know 🤮 ) to make this package browser-friendly? 

Closes #432

### DIFF
--- a/packages/log/src/log_helpers.ts
+++ b/packages/log/src/log_helpers.ts
@@ -24,7 +24,7 @@ export function truncate(str: string, maxLength: number, suffix = '...[truncated
  * Gets log level from env variable. Both integers and strings (WARNING) are supported.
  */
 export function getLevelFromEnv(): number {
-    const envVar = process.env[APIFY_ENV_VARS.LOG_LEVEL];
+    const envVar = globalThis.process?.env[APIFY_ENV_VARS.LOG_LEVEL];
 
     if (!envVar) return LogLevel.INFO;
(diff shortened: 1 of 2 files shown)


vladfrangu:
We should be providing an IIFE bundle instead through tsup 👀

barjin:
Ah, sweet - TIL... although I cannot find much about details of this :)

is [this PR](https://github.com/apify/apify-shared-js/pull/430) enough? is the explicit check for `process` existence still needed? Is there anywhere to find more about this node.js / browser development?

vladfrangu:
That PR is a start, we'll need to tweak the settings for /log to support iife but first we should get that in

PR #438, opened 2024-01-21
feat: log error causes (without breaking things this time)

vladfrangu:
This is the proper reimplementation of https://github.com/apify/apify-shared-js/pull/428 & revert of https://github.com/apify/apify-shared-js/pull/434

image

```ts
import log from '@apify/log';

const err = new Error('Some error');
const errWithCause = new Error('Some error with cause', { cause: err });
const errWithNestedCause = new Error('Some error with nested cause', { cause: errWithCause });

log.exception(errWithNestedCause);
```

### DIFF
--- a/test/log.test.ts
+++ b/test/log.test.ts
@@ -190,5 +190,19 @@ describe('log', () => {
 
             expect(line).toMatch(new RegExp(pattern));
         });
+
+        it('should support printing cause even if it is not an error', () => {
+            const log = new Log({ logger: new LoggerText() });
+            const actualError = new Error('some error', { cause: 'hello world!' });
+
+            log.exception(actualError, 'Some error message');
+
+            expect(loggerSpy).toHaveBeenCalled();
+
+            const line = loggedLines.error;
+            const pattern = `^ERROR Some error message\\s+some error([\\s\\S]*)CAUSE: hello world!`;
+
+            expect(line).toMatch(new RegExp(pattern));
+        });
     }
 });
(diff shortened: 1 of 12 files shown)


barjin:
I suppose the screenshot isn't directly 1:1 from the code, right? The error messages don't match 😅

vladfrangu:
> I suppose the screenshot isn't directly 1:1 from the code, right? The error messages don't match 😅

I deleted the script before pushing to the branch 😢, but the only difference between the code and the screenshot is the error message (the rest is the same)

test/log.test.ts
```diff
@@ -150,4 +173,22 @@ describe('log', () => {
         log.info('Something to be informed about happened', { hotel: 'restaurant' });
         expect(loggerSpy).toBeCalledWith(LEVELS.INFO, 'Something to be informed about happened', { foo: 'bar', hotel: 'restaurant' }, undefined, { prefix: null, suffix: null });
     });
+
+    // Only Node16+ supports cause
+    if (!process.version.startsWith('v14')) {
```
barjin, line 178: kinda weird to see this, isn't it safe to remove this (the more so that this is a test?)

---
edit: now I see we're testing for Node 14 in CI 👀 is there a reason for that?

barjin (approved):
lgtm otherwise :)

test/log.test.ts
vladfrangu: I already tried to nuke v14 tests in CI, but unfortunately we can't until the meteor migration is fully done 😢

PR #448, opened 2024-02-16
fix: make input_schema fully ESM usable 😢

vladfrangu:
Don't you love CJS/ESM state in nodejs


fnesveda:
I feel your pain, just came across this while migrating the worker to ESM ❤️

PR #449, opened 2024-02-20
feat: Remove @apify/consts private package dependency

jancurn:
@apify/consts is a private package, and thus the @apify/log package couldn't be used in public packages. This fixes it. See also https://apifier.slack.com/archives/C01VBUV81UZ/p1708427261484589

### DIFF
--- a/packages/log/package.json
+++ b/packages/log/package.json
@@ -48,7 +48,6 @@
         "access": "public"
     },
     "dependencies": {
-        "@apify/consts": "^2.26.0",
         "ansi-colors": "^4.1.1"
     }
 }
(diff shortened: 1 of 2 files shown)


B4nan:
all packages in this repository are public, sounds like you are confusing the private packages used in apify-core published to GH registry, everything you see here is published directly to NPM

jancurn:
@B4nan dam it, you're right, I should probably do some other job 🙈

B4nan:
with that said, i never understood why we have two sets of the packages, i ges that some needs to be private, but why are we duplicating stuff like the logger or the consts? i would say we should have only the extra private packages there and otherwise use the public ones (and for private consts we should have something like "private-consts" package)

jancurn:
🤷‍♂️

jirimoravcik:
> with that said, i never understood why we have two sets of the packages, i ges that some needs to be private, but why are we duplicating stuff like the logger or the consts? i would say we should have only the extra private packages there and otherwise use the public ones (and for private consts we should have something like "private-consts" package)

well some constants are indeed, private. and it's easier to change the private package because they're in a monorepo with other apps, so I guess that's why some things that should have ended up here are in private consts...

PR #466, opened 2024-07-17
feat: add input schema type

stetizu1:
I created an `InputSchema` type, mainly based on our documentation and also usage and tests. I am not sure if there is anything else that could be present or if `$schema` also should have been typed.

I tried to stay as backward compatible as possible.

### DIFF
--- a/packages/input_schema/src/input_schema.ts
+++ b/packages/input_schema/src/input_schema.ts
@@ -1,8 +1,16 @@
-import Ajv, { ErrorObject } from 'ajv';
+import Ajv, { ErrorObject, Schema } from 'ajv';
 import schema from './schema.json';
 import { m } from './intl';
+import {
+    FieldDefinition,
+    FieldDefinitionUnchecked,
+    InputSchema,
+    InputSchemaBaseChecked,
+    InputSchemaUnchecked, StringFieldDefinition,
+} from './types';
 
 export { schema as inputSchema };
+
 const { definitions } = schema;
 
 /**
@@ -56,9 +64,9 @@ export function parseAjvError(
 }
 
 /**
- * Validates given object against schema and throws a human readable error.
+ * Validates given object against schema and throws a human-readable error.
  */
-const validateAgainstSchemaOrThrow = <T> (validator: Ajv, obj: T, inputSchema: any, rootName: string) => {
+const validateAgainstSchemaOrThrow = (validator: Ajv, obj: InputSchemaUnchecked | FieldDefinitionUnchecked, inputSchema: Schema, rootName: string) => {
     if (validator.validate(inputSchema, obj)) return;
 
     const errorMessage = parseAjvError(validator.errors![0], rootName)?.message;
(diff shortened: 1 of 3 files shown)


packages/input_schema/src/input_schema.ts
```diff
@@ -101,9 +109,9 @@ const validateField = <T extends Record<string, any>> (validator: Ajv, fieldSche
         return;
     }
 
-    // If there are more matching definitions then the type is string and we need to get the right one.
+    // If there are more matching definitions then the type is string, and we need to get the right one.
     // If the definition contains "enum" property then it's enum type.
-    if (fieldSchema.enum) {
+    if ((fieldSchema as StringFieldDefinition).enum) {
```
stetizu1, line 114: I was a bit indecisive here.

The right thing to do is to use discriminating union
```
if (fieldSchema.type === 'string' &&
```
but that could create a breaking change

packages/input_schema/src/types.ts
```diff
@@ -0,0 +1,116 @@
...
+    | IntegerFieldDefinition
+    | ObjectFieldDefinition
+    | ArrayFieldDefinition
+    | MixedFieldDefinition
+
+// should not be present, this type is for invalid schema
+type NeverFieldDefinition = CommonFieldDefinition<never> & { type?: undefined, editor?: string }
+
+type FieldDefinitionToUnchecked<T extends FieldDefinition | NeverFieldDefinition> = Partial<
+    Omit<T, 'type'> & { type: string | string[] | readonly string[] }
```
stetizu1, line 78: I kind of hoped that we could keep discriminating union there - but... 😞 
it would break all the existing things 😅 Mainly because we are not freezing the config objects (`as const`) so they are inferred as `string`, not constants (`'string', 'number', 'array'`, etc.)

packages/input_schema/src/input_schema.ts
```diff
@@ -138,16 +153,16 @@ export function validateExistenceOfRequiredFields<T extends Record<string, any>>
  *
  * This way we get the most accurate error message for user.
  */
-export function validateInputSchema<T extends Record<string, any>>(validator: Ajv, inputSchema: T) {
+export function validateInputSchema(validator: Ajv, inputSchema: InputSchemaUnchecked): asserts inputSchema is InputSchema {
```
fnesveda, line 156: Is the signature change here correct? I think you should be able to pass any object as `inputSchema` value, otherwise this would be a breaking change.

fnesveda (changes requested):
Cool! I'm not sure if the change of the `inputSchema` argument to `InputSchemaUnchecked` is correct, though.

Also, this should definitely be a `feat`, not just `chore`.

Also adding @B4nan as reviewer to check if this will have any implications for the SDK or CLI.

packages/input_schema/src/input_schema.ts
stetizu1: Yeah, now when I think about it....
Yes, I thought that the deep partial type would be enough, but you are correct, since type would fail if you pass for example `description: 3`

stetizu1: I changed it so unchecked type is only one for middle state and everything else is `Record<string, unknown>` or `Record<string<Record<string, unknown>>`

packages/input_schema/src/types.ts
```diff
@@ -0,0 +1,93 @@
+type CommonFieldDefinition<T> = {
+    title: string;
+    description: string;
+    default?: T;
+    prefill?: T extends boolean ? never : T;
```
B4nan, line 5: didn't we add support for boolean prefills recently? found this PR as well as some messages on slack on this topic

https://github.com/apify/apify-core/pull/16390

B4nan (approved):
few nits, nothing blocking

> Also adding @B4nan as reviewer to check if this will have any implications for the SDK or CLI.

on first look it should be fine, it shouldn't break anything for downstream, worst case it will produce some type issues, but those are not relevant for the users (and can be addressed when upgrading the dependency in the repo).

fnesveda (approved):
Cool!

packages/input_schema/src/types.ts
stetizu1: Ok, maybe it should be added to our docs. I will remove it there. Same with docs and editor: 'json' option in string type (it is currently missing there, even though we maybe accept it?). Maybe @TC-MO ?

packages/input_schema/src/types.ts
```diff
@@ -0,0 +1,93 @@
...
+    example?: T;
+    nullable?: boolean;
+    sectionCaption?: string;
+    sectionDescription?: string;
+}
+
+export type StringFieldDefinition = CommonFieldDefinition<string> & { type: 'string' } & {
```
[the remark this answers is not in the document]
stetizu1: It made more sense when the rest was partial 😅 I moved it.

packages/input_schema/src/types.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+type AllTypes = StringFieldDefinition['type']
+    | BooleanFieldDefinition['type']
+    | IntegerFieldDefinition['type']
+    | ObjectFieldDefinition['type']
+    | ArrayFieldDefinition['type']
+
+export type MixedFieldDefinition = CommonFieldDefinition<never> & {
+    type: AllTypes[] | readonly AllTypes[];
```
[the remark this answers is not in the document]
stetizu1: Yes, sure. And there were actually multiple arrays that needed readonly so they could accept the freezed object, that I missed. So thanks for the catch! I also fixed the $schema and required to be optional on the object 😅

PR #468, opened 2024-08-01
fix(markdown): upgrade dependency on `marked` to v13

B4nan:
**BREAKING CHANGE:**
Node 18 is required for `@apify/markdown` package, since `marked` added this constraint too via `engines` field. The tests are still passing even with node 14, so this is a rather artificial constraint.

Signature of `customHeadingRenderer` changed:

```diff
-customHeadingRenderer(text: string, level: 1 | 2 | 3 | 4 | 5 | 6, raw: string): string
+customHeadingRenderer({ depth, text, raw }: Tokens.Heading): string
```


janbuchar (approved):
LGTM

PR #471, opened 2024-08-01
chore: remove `@apify/salesforce_client` package

B4nan:
This is no longer used anywhere.

Also removes dev dependency on `deploy-web-to-s3` which wasn't used anywhere and had some security issues in its deps.


vladfrangu (approved):
Lets mark this as deprecated on npm!

PR #477, opened 2024-09-16
feat(input_schema): datepicker allowAbsolute/allowRelative properties and validation

MFori:
This PR introduce new input_schema properties `allowAbsolute` and `allowRelative` used when the `editor` property is set to `datepicker`:
- the `schema.json` is updated in the way, that properties `allowAbsolute` and `allowRelative` are valid only when `editor` is `datepicker`, `allowAbsolute` is considered as default and any of `allowAbsolute` or `allowRelative` is true
- value format validation
- unit tests for schema validation and value validation
- New datepicker with support for absolute and relative dates PR: https://github.com/apify/apify-core/pull/17465
- Docs PR: https://github.com/apify/apify-docs/pull/1227

### DIFF
--- a/packages/input_schema/src/types.ts
+++ b/packages/input_schema/src/types.ts
@@ -18,6 +18,9 @@ export type StringFieldDefinition = CommonFieldDefinition<string> & {
     enum?: readonly string[]; // required if editor is 'select'
     enumTitles?: readonly string[]
     isSecret?: boolean;
+    // used for 'datepicker' editor, isAbsolute is considered with default value true
+    isAbsolute?: boolean;
+    isRelative?: boolean;
 }
 
 export type BooleanFieldDefinition = CommonFieldDefinition<boolean> & {
--- a/packages/input_schema/src/types.ts
+++ b/packages/input_schema/src/types.ts
@@ -18,9 +18,10 @@ export type StringFieldDefinition = CommonFieldDefinition<string> & {
     enum?: readonly string[]; // required if editor is 'select'
     enumTitles?: readonly string[]
     isSecret?: boolean;
-    // used for 'datepicker' editor, isAbsolute is considered with default value true
-    isAbsolute?: boolean;
-    isRelative?: boolean;
+    // Used for 'datepicker' editor, allowAbsolute is considered with default value true
+    // If only relative time is wanted, allowAbsolute must be explicitly set to false
+    allowAbsolute?: boolean;
+    allowRelative?: boolean;
 }
 
 export type BooleanFieldDefinition = CommonFieldDefinition<boolean> & {
--- a/packages/input_schema/src/utilities.ts
+++ b/packages/input_schema/src/utilities.ts
@@ -253,7 +253,7 @@ export function validateInputUsingValidator(
             const acceptAbsolute = isAbsolute !== false;
             const acceptRelative = isRelative === true;
             const isValidAbsolute = /^[0-9]{4}-[0-9]{2}-[0-9]{2}$/.test(value);
-            const isValidRelative = /^[+-] [0-9]+ (day|month|year)s?$/.test(value);
+            const isValidRelative = /^[+-] [0-9]+ (day|week|month|year)s?$/.test(value);
             let isValidDate: boolean | undefined;
 
             if (isValidAbsolute) {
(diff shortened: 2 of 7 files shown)


packages/input_schema/src/utilities.ts
```diff
@@ -247,6 +247,50 @@ export function validateInputUsingValidator(
                 }
             }
         }
+
+        // Check datepicker editor format
+        if (type === 'string' && editor === 'datepicker' && value && typeof value === 'string') {
+            const acceptAbsolute = isAbsolute !== false;
+            const acceptRelative = isRelative === true;
+            const isValidAbsolute = /^[0-9]{4}-[0-9]{2}-[0-9]{2}$/.test(value);
+            const isValidRelative = /^[+-] [0-9]+ (day|week|month|year)s?$/.test(value);
```
drobnikj, line 256: The docs do not specify the unit. What about minutes and seconds?

drobnikj: I got it, it is date picker, so the minutes and seconds are not valid, so please add this to docs, and allow units. I will add note in doc PR as well.

packages/input_schema/src/types.ts
```diff
@@ -18,6 +18,9 @@ export type StringFieldDefinition = CommonFieldDefinition<string> & {
     enum?: readonly string[]; // required if editor is 'select'
     enumTitles?: readonly string[]
     isSecret?: boolean;
+    // used for 'datepicker' editor, isAbsolute is considered with default value true
+    isAbsolute?: boolean;
```
drobnikj, line 22: Why do we have two fields isAbsolute and isRelative, is it isRelative enaght?
if the field is isRelative it is relative otherwise it is absolute, the absolute is default value anyway

packages/input_schema/src/utilities.ts
```diff
@@ -247,6 +247,50 @@ export function validateInputUsingValidator(
                 }
             }
         }
+
+        // Check datepicker editor format
+        if (type === 'string' && editor === 'datepicker' && value && typeof value === 'string') {
```
drobnikj, line 252: If you have just isRelative the validation here will be much easier :)

drobnikj (changes requested):
This package needs to be heavily unit tested and I see no tests here. Please do unit tests for your changes.

packages/input_schema/src/types.ts
MFori: Because it can accept both formats, so if only `isRelative: true` is specified it accepts both unless `isAbsolute` is explicitly set to `false`

packages/input_schema/src/utilities.ts
MFori: Added to docs PR

packages/input_schema/src/types.ts
drobnikj: Using better naming for param isRelative and isAbsolute are confusing, allowAbsolute and allowRelative?

test/input_schema_definition.test.ts
```diff
@@ -285,5 +285,163 @@ describe('input_schema.json', () => {
...
+                    properties: {
+                        myField: {
+                            title: 'Field title',
+                            description: 'My test field',
+                            type: 'string',
+                            editor: 'datepicker',
+                            allowAbsolute: false,
+                        },
+                    },
+                })).toBe(false);
```
drobnikj, line 425: I would rather test if the error message is correct as the logic is not trivial.

test/input_schema_definition.test.ts
```diff
@@ -285,5 +285,163 @@ describe('input_schema.json', () => {
...
+                        myField: {
+                            title: 'Field title',
+                            description: 'My test field',
+                            type: 'string',
+                            editor: 'datepicker',
+                            allowAbsolute: false,
+                            allowRelative: false,
+                        },
+                    },
+                })).toBe(false);
```
drobnikj, line 443: Same here.

drobnikj (changes requested):
👏  for tests, let's fix last bits and we are good to go.

packages/input_schema/src/utilities.ts
```diff
@@ -137,12 +137,12 @@ export function validateInputUsingValidator(
     }
 
     Object.keys(properties).forEach((property) => {
-        const value = input[property] as Record<string, any>;
-        const { type, editor, patternKey, patternValue } = properties[property];
+        const value = input[property] as Record<string, any> | Array<any> | string | number | boolean | null;
```
B4nan, line 140: I wonder if you really need this long type hint, maybe `unknown` would be just fine, since you have all the type guards anyway.

B4nan (approved):
only had a brief look

test/input_schema_definition.test.ts
MFori: Added check to error messages

drobnikj (approved):
Nice 👍

PR #480, opened 2024-10-08
feat: add max cost per run 

novotnyj:
Needed to add `maxCostPerRunUsd` from run options to actor env vars.


jirimoravcik:
guys, just FYI
this needs to be added to https://github.com/apify/actor-whitepaper?tab=readme-ov-file#environment-variables cc @jancurn 
also https://docs.apify.com/platform/actors/development/programming-interface/environment-variables

jirimoravcik (approved):

novotnyj:
@mhamas Looking at https://github.com/apify/actor-whitepaper?tab=readme-ov-file#environment-variables - maybe we should use `ACTOR_MAX_TOTAL_CHARGE_USD` for this?

jancurn:
Yes, I called this `ACTOR_MAX_TOTAL_CHARGE_USD` because the function is called `Actor.charge()`, for consistency. Also, the word "cost" is used on platform in the context of platform costs, which is typically lower than what is charged, because devs have margin. From point of view of the developer, "costs" is their costs, not what they charge their users.

mhamas (approved):
Discussed with @novotnyj and we can rename this on platform as well. @novotnyj can you create a ticket for this into this sprint and do it with higher priority, co we can get it done before it actually starts to be properly used. Thx

novotnyj:
https://github.com/apify/apify-core/issues/17855

PR #481, opened 2024-10-09
chore: update dependencies

B4nan:


vladfrangu (approved):
Well thats a lot of removals from lockfile 👀🫣

B4nan:
well, but it still works :D i think its about the outdated lockfile we had here, mine is v3 but the one in master is v2. i guess i should also [bump this](https://github.com/apify/apify-shared-js/blob/master/renovate.json#L14).

PR #483, opened 2024-10-15
fix(input_schema): disable isSecret for array properties

MFori:
Solves: https://github.com/apify/apify-core/issues/15613
The issue was with `arrayProperty` because it enables additional properties (`"additionalProperties": true`) so not defined properties were accepted including `isSecret`.

This PR fixes `arrayProperty` ensuring that not defined properties are prohibited

Adding also missing test that check this


gippy (approved):

drobnikj (approved):
🤞

PR #484, opened 2024-10-22
feat(input_schema): Resource properties

MFori:
https://github.com/apify/apify-core/issues/12871

TODO:
- [ ] write tests

### DIFF
--- a/packages/input_schema/src/schema.json
+++ b/packages/input_schema/src/schema.json
@@ -39,6 +39,8 @@
                         { "$ref": "#/definitions/objectProperty" },
                         { "$ref": "#/definitions/integerProperty" },
                         { "$ref": "#/definitions/booleanProperty" },
+                        { "$ref": "#/definitions/resourceProperty" },
+                        { "$ref": "#/definitions/resourceArrayProperty" },
                         { "$ref": "#/definitions/anyProperty" }
                     ]
                 }
@@ -86,7 +88,7 @@
                 "title": { "type": "string" },
                 "description": { "type": "string" },
                 "nullable": { "type": "boolean" },
-                "editor": { "enum": ["javascript", "python", "textfield", "textarea", "datepicker", "hidden", "dataset", "keyValueStore", "requestQueue"] },
+                "editor": { "enum": ["javascript", "python", "textfield", "textarea", "datepicker", "hidden"] },
                 "isSecret": { "type": "boolean" }
             },
             "required": ["type", "title", "description", "editor"],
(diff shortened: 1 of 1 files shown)


packages/input_schema/src/schema.json
```diff
@@ -32,22 +32,80 @@
...
-                        { "$ref": "#/definitions/arrayProperty" },
-                        { "$ref": "#/definitions/objectProperty" },
-                        { "$ref": "#/definitions/integerProperty" },
-                        { "$ref": "#/definitions/booleanProperty" },
-                        { "$ref": "#/definitions/anyProperty" }
-                    ]
+                    "if": {
+                        "properties": {
+                            "resourceType": {
+                                "not": {
```
drobnikj, line 38: Nit: why not in condition? I prefer condition without not and switch else branches, it will make it more readable.

packages/input_schema/src/schema.json
```diff
@@ -32,22 +32,80 @@
...
+                    }
                 }
             }
         }
     },
     "additionalProperties": false,
     "required": ["title", "type", "properties", "schemaVersion"],
     "definitions": {
+        "resourceProperty": {
+            "title": "Resource property",
```
drobnikj, line 69: Nit: It looks the title, description, nullable, sectionCaption, sectionDescription appear in all definition. Could we move these to special definitions? 
```
{
...
    "definitions": {
        "baseProperty": {
            "type": "object",
            "properties": {
                "title": {"type": "string"},
                "description": {
                    "type": "string"
                },
                "nullable": { "type": "boolean"}
            },
            "required": ["title", "description"]
        },
        "resourceProperty": {
            "allOf": [
                {"$ref": "#/definitions/baseProperty"},
                {
                    "properties": {
                        ...the rest of props on top of baseProperty
                    }
                }
            ],
            "required": ["type", "title", "description", "resourceType"]
        },
        ...
    }
...
}
```

drobnikj (commented):
Looks good, just have two idea on how we can simplify.

gippy (approved):
I like the "allOf" base property Kuba has. Otherwise preapproving

MFori:
The solution with baseProperty and `allOf` has some issues:
- `additionalProperties` doesn't work for this, but yeah we can use `unevaluatedProperties` in this case
- we are accessing field `property.properties` in our custom validation and this property wouldn't have this property and additional transformation will be needed to check if the property have this field or not and then merge properties from the `allOf` items.
- the `baseProperty` is still validated and will be need to filter it out in our custom validation
- lot of tests fails also on other issues then the points above and needs to be resolved

So it's possible but do we really need it in this case with the increased complexity of validation?
@gippy @drobnikj 

According to: `why not in condition? I prefer condition without not and switch else branches, it will make it more readable.`
If I switch the else branches lot of our tests fails I am on it and trying to figure out what's going on there

gippy:
> The solution with baseProperty and `allOf` has some issues:
> 
> * `additionalProperties` doesn't work for this, but yeah we can use `unevaluatedProperties` in this case
> * we are accessing field `property.properties` in our custom validation and this property wouldn't have this property and additional transformation will be needed to check if the property have this field or not and then merge properties from the `allOf` items.
> * the `baseProperty` is still validated and will be need to filter it out in our custom validation
> * lot of tests fails also on other issues then the points above and needs to be resolved
> 
> So it's possible but do we really need it in this case with the increased complexity of validation? @gippy @drobnikj
> 
> According to: `why not in condition? I prefer condition without not and switch else branches, it will make it more readable.` If I switch the else branches lot of our tests fails I am on it and trying to figure out what's going on there

I liked the idea because it reduces repetition of the schema, but if it's at the cost of increased complexity of the schema, then I do not mind the current solution where the properties are repeated.

MFori:
> Nit: why not in condition? I prefer condition without not and switch else branches, it will make it more readable.

@drobnikj I was able to remove this condition at all by updating our custom validation and moving some tests from `input_schema_definition.test.ts` to `input_schema.test.ts` because the second one uses our validation but the first one doesn't.

packages/input_schema/src/types.ts
```diff
@@ -61,6 +61,19 @@ export type ArrayFieldDefinition = CommonFieldDefinition<Array<unknown>> & {
...
+export type CommonResourceFieldDefinition<T> = CommonFieldDefinition<T> & {
+    editor?: 'resourcePicker' | 'hidden';
+    resourceType: 'dataset' | 'keyValueStore' | 'requestQueue'
+}
+
+export type ResourceFieldDefinition = CommonResourceFieldDefinition<string> & {
+    type: 'string'
+}
+
+export type ResourceArrayFieldDefinition = CommonResourceFieldDefinition<Array<string>> & {
```
B4nan, line 73: i dont think we use `Array<T>` anywhere else, lets stay consistent

```suggestion
export type ResourceArrayFieldDefinition = CommonResourceFieldDefinition<string[]> & {
```

B4nan (approved):

drobnikj (approved):

PR #485, opened 2024-11-01
feat(input_schema): datepicker dateType property

MFori:
This PR replace `allowAbsolute`/`allowRelative` boolean properties for datepicker with single `dateType` property that accepts values `absolute`, `relative` and `absoluteOrRelative` based on:
- https://github.com/apify/apify-docs/pull/1227
- https://apify.slack.com/archives/C010Q0FBYG3/p1730210649660309?thread_ts=1728907212.870409&cid=C010Q0FBYG3

`dateType` property:
- is optional
- is valid only when `editor` is `datepicker`
- is only for ui input representation, no validation is executed based on this property


packages/input_schema/src/schema.json
```diff
@@ -149,8 +139,8 @@
                         "nullable": { "type": "boolean" },
                         "minLength": { "type": "integer" },
                         "maxLength": { "type": "integer" },
-                        "editor": { "enum": ["javascript", "python", "textfield", "textarea", "datepicker", "hidden"] },
-                        "isSecret": { "type": "boolean" },
+                        "editor": { "enum": ["javascript", "python", "textfield", "textarea", "hidden"] },
+                        "isSecret": { "enum": [false] },
```
gippy, line 143: This is to force it to just a single value?

gippy (approved):

fnesveda (approved):
As I understand this, this is a breaking change, because suddenly the input schemas with the old `allowAbsolute` etc. properties will be invalid. Just to be sure, is that OK? Were they not documented anywhere yet? Are there no Actors with input schemas containing those properties?

Otherwise looks good.

MFori:
> As I understand this, this is a breaking change, because suddenly the input schemas with the old `allowAbsolute` etc. properties will be invalid. Just to be sure, is that OK? Were they not documented anywhere yet? Are there no Actors with input schemas containing those properties?
> 
> Otherwise looks good.

Yes, these old values `allowAbsolute` and `allowRelative` were never documented anywhere and are not used by any Actors.

PR #486, opened 2024-11-04
feat(consts): introduce profile name (first, last, organization, ...) restrictions and adjust username regex

DaveHanns:
Affects https://github.com/apify/apify-core/issues/17545

Besides introduction of new Profile name restrictions, this also adjusts restrictions (regex) of valid username.
It does so to prevent users from inputing viable url addresses by removing a dot from allowed characters. By doing so, the username now cannot form a valid URL or Email domain.
See comment bellow for more info (@baldasseva mainly).

### DIFF
--- a/packages/consts/src/consts.ts
+++ b/packages/consts/src/consts.ts
@@ -134,7 +134,7 @@ export const PROFILE_NAME = {
     MAX_LENGTH: 50,
 
     // Regex to prohibit anything that could potentially form an URL or email address.
-    REGEX: /^[^/.@><]{3,50}$/,
+    REGEX: /^[^/.@><]*$/,
 };
 
 /**
--- a/packages/consts/src/consts.ts
+++ b/packages/consts/src/consts.ts
@@ -116,7 +116,7 @@ export const USERNAME = {
 
     // Regex matching a potentially allowed username. The numbers must match MIN and MAX!
     // Note that username must also pass isForbiddenUser() test to be allowed!
-    REGEX: /^[a-zA-Z0-9_-]{3,30}$/,
+    REGEX: /^[a-zA-Z0-9_.-]{3,30}$/,
 };
 
 export const EMAIL = {
@@ -133,8 +133,8 @@ export const PROFILE_NAME = {
     MIN_LENGTH: 3,
     MAX_LENGTH: 50,
 
-    // Regex to prohibit anything that could potentially form an URL or email address.
-    REGEX: /^[^/.@><]*$/,
+    // Prohibits usage of @, <, > and :// in the name
+    REGEX: /^(?!.*:\/\/)[^@><]*$/,
 };
 
 /**
--- a/packages/consts/src/consts.ts
+++ b/packages/consts/src/consts.ts
@@ -1,4 +1,4 @@
-import { DNS_SAFE_NAME_REGEX } from './regexs';
+import { DNS_SAFE_NAME_REGEX, EMAIL_REGEX } from './regexs';
 
 export const FREE_SUBSCRIPTION_PLAN_CODE = 'DEV';
 
@@ -117,6 +117,13 @@ export const USERNAME = {
     // Regex matching a potentially allowed username. The numbers must match MIN and MAX!
     // Note that username must also pass isForbiddenUser() test to be allowed!
     REGEX: /^[a-zA-Z0-9_-]{3,30}$/,
+};
+
+export const EMAIL = {
+    MIN_LENGTH: 3,
+    MAX_LENGTH: 254, // see https://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1690
+
+    REGEX: EMAIL_REGEX,
 };
 
 /**
--- a/packages/consts/src/consts.ts
+++ b/packages/consts/src/consts.ts
@@ -98,6 +98,11 @@ export const ACTOR_TYPES = {
 } as const;
 
 /**
+ * Used as username for returning user own info from API v2/users/username
+ */
+export const ME_USER_NAME_PLACEHOLDER = 'me';
+
+/**
  * Username used when user is anonymous.
  */
 export const ANONYMOUS_USERNAME = 'anonymous';
@@ -111,7 +116,18 @@ export const USERNAME = {
 
     // Regex matching a potentially allowed username. The numbers must match MIN and MAX!
     // Note that username must also pass isForbiddenUser() test to be allowed!
-    REGEX: /^[a-zA-Z0-9_.-]{3,30}$/,
+    REGEX: /^[a-zA-Z0-9_-]{3,30}$/,
+};
+
+/**
+ * Profile name (such as organization or first / last name) constraints.
+ */
+export const PROFILE_NAME = {
+    MIN_LENGTH: 3,
+    MAX_LENGTH: 50,
+
+    // Regex to prohibit anything that could potentially form an URL or email address.
+    REGEX: /^[^/.@><]{3,50}$/,
 };
 
 /**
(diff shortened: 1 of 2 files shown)


packages/consts/src/consts.ts
```diff
@@ -111,7 +116,25 @@ export const USERNAME = {
 
     // Regex matching a potentially allowed username. The numbers must match MIN and MAX!
     // Note that username must also pass isForbiddenUser() test to be allowed!
-    REGEX: /^[a-zA-Z0-9_.-]{3,30}$/,
+    REGEX: /^[a-zA-Z0-9_-]{3,30}$/,
+};
+
+export const EMAIL = {
+    MIN_LENGTH: 3,
+    MAX_LENGTH: 254, // see https://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1690
```
DaveHanns, line 124: Moved from apify-core as we had quite a bit of inconsistencies in a ways we handle email restrictions across the org.

packages/consts/src/consts.ts
```diff
@@ -111,7 +116,25 @@ export const USERNAME = {
 
     // Regex matching a potentially allowed username. The numbers must match MIN and MAX!
     // Note that username must also pass isForbiddenUser() test to be allowed!
-    REGEX: /^[a-zA-Z0-9_.-]{3,30}$/,
+    REGEX: /^[a-zA-Z0-9_-]{3,30}$/,
```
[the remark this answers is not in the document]
DaveHanns: For a clarification, excluding the dot makes this regex effectively exclude any viable URLs and email addresses.

test/consts.test.ts
```diff
@@ -6,12 +6,9 @@ describe('consts', () => {
         it('REGEX works as expected', () => {
             expect(USERNAME.REGEX.test('anonymous')).toBe(true);
             expect(USERNAME.REGEX.test('---')).toBe(true);
-            expect(USERNAME.REGEX.test('john.doe')).toBe(true);
```
B4nan, line 9: instead of removing tests, you should just change the assertions

packages/consts/src/consts.ts
```diff
@@ -111,7 +116,25 @@ export const USERNAME = {
...
+
+/**
+ * Profile name (such as organization or first / last name) constraints.
+ */
+export const PROFILE_NAME = {
+    MIN_LENGTH: 3,
+    MAX_LENGTH: 50,
+
+    // Regex to prohibit anything that could potentially form an URL or email address.
+    REGEX: /^[^/.@><]*$/,
```
DaveHanns, line 137: The originally proposed solution of excluding strings with '//:' in them did not work completely. Gmail and other email providers (tested also using Seznam and Yahoo) recognized also urls without the protocol and emails without the emailto protocol.

Excluding the dot, slash and @ effectively excludes any viable urls and email addresses. The > and < characters were previously checked and prohibited in custom function in apify-core (not sure what was the reason, probably something interpolation-wise 🤷).

test/consts.test.ts
DaveHanns: Agree. Will fix.

packages/consts/src/consts.ts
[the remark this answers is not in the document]
valekjo: We have quite a few users who have domains as usernames (the definitely valid ones are usually the domain of the company).

And it seems that profile.name is a similar case.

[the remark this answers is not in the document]
DaveHanns: As it will be implemented in the apify-core (see draft [here](https://github.com/apify/apify-core/pull/18188)), existing users (nor their usernames nor profile names) will not be affected till they decide to modify their data.
It should affect them only when attempt to create new user (segmentation) / organization, update username or profile and to convert user to organization is made.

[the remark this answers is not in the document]
baldasseva: Thanks for the comment. I checked on the web and we use the function in two kinds of pages: actor detail (e.g. https://apify.com/trudax/reddit-scraper) and user detail (e.g. https://apify.com/trudax) => when there is the username in the url.
If the username is deemed "not valid", we send back a 404 Not Found.

So the problem should be clear: if we have any users (and their actors) that use the dot, we will give an error on their page even if their username was valid at the time when they chose it. From a quick look to the store I see that a lot of users have dots in their usernames

So my question is: is any action going to be taken for the users that already have the bad usernames? If not, then I think it's acceptable that in console we keep using the old regex, since we only have the two usecases mentioned above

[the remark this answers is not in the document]
valekjo: Personally, I'd keep the dot, and just deal with "proper" links.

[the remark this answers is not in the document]
DaveHanns: Yes, that would be a viable solution for a web (as the regex there is not used for any mutations).

If we are willing to live with some links (urls) getting through (and therefore being rendered as links by Slack, Gmail and such), I have no problem with keeping the dot in both username and profile names (org, first, last, ...) as suggested by @valekjo.

@gippy, what do you think ?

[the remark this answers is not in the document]
gippy: I believe the dot check is too strict and it should be enough to prevent users from intentionally writing URLs. If a email client converts the name to url because it contains `.` it's outside of our control.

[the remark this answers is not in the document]
DaveHanns: @baldasseva, with this decision, no need for any change in apify-web.

packages/consts/src/consts.ts
```diff
@@ -114,6 +119,19 @@ export const USERNAME = {
     REGEX: /^[a-zA-Z0-9_.-]{3,30}$/,
```
baldasseva, line 119: You cannot form urls with protocol (like http://) because : and // are not allowed. But what about `www.`? Are we fine with that one? (Like you mentioned [here](https://github.com/apify/apify-shared-js/pull/486#discussion_r1827339816))

DaveHanns: Yeah, we agreed that it's not worth the hustle and would be too strict, as there are a lot of users / orgs named something.io, else.com, ...

It also means, that the apify-web won't be affected in any way.

test/consts.test.ts
DaveHanns: Fixed.

terezavrchovinova (approved):
LGTM, the only thing I've noticed is that you can still put "mailto://" into the Discord user ID. I don't think it was in the scope of this task to change it, so pre-approving.

Screenshot 2024-11-05 at 14 55 36

test/consts.test.ts
```diff
@@ -17,6 +27,30 @@ describe('consts', () => {
...
+            expect(PROFILE_NAME.REGEX.test('Simple:.//Name')).toBe(true);
+            expect(PROFILE_NAME.REGEX.test('Joh////:/n-Doe')).toBe(true);
+            expect(PROFILE_NAME.REGEX.test('user:name')).toBe(true);
+
+            // Invalid cases
+            expect(PROFILE_NAME.REGEX.test('user@name')).toBe(false);
+            expect(PROFILE_NAME.REGEX.test('user>name')).toBe(false);
+            expect(PROFILE_NAME.REGEX.test('user<name')).toBe(false);
+            expect(PROFILE_NAME.REGEX.test('example://test')).toBe(false);
+            expect(PROFILE_NAME.REGEX.test('example://////test')).toBe(false);
```
valekjo, line 50: Nit: I'd include examples that showcase why we want such regex. `<a href="https://malicious.site">Safe John Doe</a>`

packages/consts/src/consts.ts
```diff
@@ -114,6 +119,19 @@ export const USERNAME = {
...
+    REGEX: EMAIL_REGEX,
+};
+
+/**
+ * Profile name (such as organization or first / last name) constraints.
+ */
+export const PROFILE_NAME = {
+    MAX_LENGTH: 50,
+    REGEX: /^(?!.*:\/\/)[^@><]*$/, // Prohibits usage of @, <, > and ://
+};
```
valekjo, line 133: Question: Do we want to have this in apify-shared, even if it's just used in apify-core?

valekjo (approved):

gippy (approved):

PR #492, opened 2024-12-02
feat(utilities): markedSetNofollowLinks with hostname check

MFori:
Needed to solve https://github.com/apify/apify-core/issues/18466.
Now the `rel="noopener noreferrer nofollow" target="_blank"` is added only if url is outside from `apify.com` domain.
This enable passing hostname to `markedSetNofollowLinks` function and setting `rel="noopener noreferrer" target="_blank"` for apify domain but different hostname.

Once this is merged, next step would be to pass hostname from `safeMarkdown` in console/frontend down to `markedSetNofollowLinks`

### DIFF
--- a/packages/utilities/src/utilities.client.ts
+++ b/packages/utilities/src/utilities.client.ts
@@ -158,9 +158,9 @@ export function normalizeUrl(url: string, keepFragment?: boolean) {
 }
 
 // Helper function for markdown rendered marked
-// If passed hostname, it renders links outside that hostname in readme with rel="noopener noreferrer" and target="_blank" attributes
+// If passed referrerHostname, it renders links outside that hostname in readme with rel="noopener noreferrer" and target="_blank" attributes
 // And links outside apify.com in readme with rel="noopener noreferrer nofollow" and target="_blank" attributes
-export function markedSetNofollowLinks(href: string, title: string, text: string, hostname?: string) {
+export function markedSetNofollowLinks(href: string, title: string, text: string, referrerHostname?: string) {
     let urlParsed: URL;
     try {
         urlParsed = new URL(href);
@@ -168,7 +168,7 @@ export function markedSetNofollowLinks(href: string, title: string, text: string
         // Probably invalid url, go on
     }
     const isApifyLink = (urlParsed! && /(\.|^)apify\.com$/i.test(urlParsed.hostname));
-    const isSameHostname = !hostname || (urlParsed! && urlParsed.hostname === hostname);
+    const isSameHostname = !referrerHostname || (urlParsed! && urlParsed.hostname === referrerHostname);
 
     if (isApifyLink && isSameHostname) {
         return `<a href="${href}">${title || text}</a>`;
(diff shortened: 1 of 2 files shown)


packages/utilities/src/utilities.client.ts
```diff
@@ -158,18 +158,25 @@ export function normalizeUrl(url: string, keepFragment?: boolean) {
 }
 
 // Helper function for markdown rendered marked
-// Renders links outside apify.com in readme with rel="noopener noreferrer nofollow" and target="_blank" attributes
-export function markedSetNofollowLinks(href: string, title: string, text: string) {
+// If passed hostname, it renders links outside that hostname in readme with rel="noopener noreferrer" and target="_blank" attributes
+// And links outside apify.com in readme with rel="noopener noreferrer nofollow" and target="_blank" attributes
+export function markedSetNofollowLinks(href: string, title: string, text: string, hostname?: string) {
```
barjin, line 163: So the new `hostname` param is the hostname of the current page (the one we're rendering the link on), right? Maybe sth like `referrerHostname` (or even passing the entire URL, not just the hostname) would help with readability a bit?

packages/utilities/src/utilities.client.ts
```diff
@@ -158,18 +158,25 @@ export function normalizeUrl(url: string, keepFragment?: boolean) {
...
         // Probably invalid url, go on
     }
     const isApifyLink = (urlParsed! && /(\.|^)apify\.com$/i.test(urlParsed.hostname));
-    return (isApifyLink)
-        ? `<a href="${href}">${title || text}</a>`
-        : `<a rel="noopener noreferrer nofollow" target="_blank" href="${href}">${title || text}</a>`;
+    const isSameHostname = !hostname || (urlParsed! && urlParsed.hostname === hostname);
+
+    if (isApifyLink && isSameHostname) {
+        return `<a href="${href}">${title || text}</a>`;
```
barjin, line 174: Out of curiosity - why don't we use the DOM API (`createElement`, `setAttribute` etc)? I'm sure this is handled in many other layers, but syntax like this always makes me think of XSS. I see the file is called `utilities.client.ts`, this means it only runs in browsers, right? (so DOM API should be always available)

barjin (commented):
@MFori I don't think our team (@B4nan or me) should have any say in this 😅 It looks safe to me, but I'd rather have somebody more web-devvy check this. Nevertheless, I left a few curious comments, feel free to ignore both :)

packages/utilities/src/utilities.client.ts
MFori: I was thinking the same at first and maybe loading the current host name just inside the function, but even it's called `utilities.client.ts` it is used in daemons and console backend 🤔

B4nan:
I rewrote the whole codebase to TS and moved it to a monorepo :)

packages/utilities/src/utilities.client.ts
barjin: Gotcha — weird indeed. Since this has worked until now and nobody complained, we don't need to rethink it right now. thanks!

barjin (approved):
Looking into this more, I'm all for merging this - now I see it just adds a new branch for Apify links from a different hostname, the rest of the code is untouched. LGTM, thanks!

drobnikj (approved):
👍

PR #500, opened 2025-02-11
feat: add create hmac signature

danpoletaev:
**This PR adds new function to the utility package.**

Function `createHmacSignature()` will be used in Apify SDK as well as in Apify Core repositories

### DIFF
--- a/packages/utilities/src/hmac.ts
+++ b/packages/utilities/src/hmac.ts
@@ -1,0 +1,20 @@
+import crypto from 'crypto';
+
+import base62 from 'base62';
+
+/**
+ * Generates an HMAC signature and encodes it using Base62.
+ * Base62 encoding reduces the signature length.
+ *
+ * @param secretKey {string} Secret key used for signing signatures
+ * @param message {string} Message to be signed
+ * @returns string
+ */
+export function createHmacSignature(secretKey: string, message: string): string {
+    const signature = crypto.createHmac('sha256', secretKey)
+        .update(message)
+        .digest('hex')
+        .substring(0, 30);
+
+    return base62.encode(parseInt(signature, 16));
+}
--- a/packages/utilities/src/hmac.ts
+++ b/packages/utilities/src/hmac.ts
@@ -1,6 +1,23 @@
 import crypto from 'crypto';
 
-import base62 from 'base62';
+const CHARSET = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');
+
+/**
+ * Encodes number to base62.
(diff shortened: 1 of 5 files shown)


B4nan:
I am not very happy to see any additional dependencies in the utilities package. We use that in many places (crawlee core and Apify CLI to begin with), and they will now all get a new dependency, even though it will be used in only the SDK and apify-core. The function seems too trivial to justify this.

I know it's really a small dependency, but at the same time the new helper is just a 5 lines of code.

fnesveda:
@B4nan is it really such an issue? I feel like you're overoptimizing, the additional complexity from sharing the helper in multiple places is not worth 0.02% of savings in Crawlee.

```bash
> npx howfat crawlee
crawlee@3.12.2 (575 deps, 95.66mb, 19921 files, ©Apache-2.0)

> npx howfat @apify/utilities
@apify/utilities@2.12.2 (3 deps, 695.66kb, 33 files, ©Apache-2.0)

> npx howfat base62
base62@2.0.2 (25.88kb, 18 files, ©MIT)
```

B4nan:
I am just pointing out a bad practice, especially when it comes to public packages. You picked the fattest library we have for your comparison.

Btw you say I am overoptimizing, I say you are over-engineering. Pick your poison :] 

Anyway, I am fine with this, I just don't like it.

danpoletaev:
The encode function is pretty simple. I copied it from https://github.com/base62/base62.js/blob/master/lib/ascii.js to avoid new dependency.

Hope everyone is ok with that

B4nan (approved):

fnesveda (changes requested):
I'm afraid this won't work correctly. if the signature is 30 hex characters, and you parse it to an integer, it's max possible value is 2^120, which is far larger than the max safe integer in JavaScript (`Number.MAX_SAFE_INTEGER` , equal to 2^53 - 1).

You'll have to use `BigInt`s for this.

packages/utilities/src/hmac.ts
```diff
@@ -0,0 +1,37 @@
+import crypto from 'crypto';
+
+const CHARSET = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');
+
+/**
+ * Encodes BigInt to base62.
+ * To avoid new dependency, this function was copied from https://github.com/base62/base62.js/blob/master/lib/ascii.js
```
fnesveda, line 7:
```suggestion
```
You can remove this comment now, since it's basically a new function 🙂 And we won't have to worry about including the license of that package.

fnesveda (approved):
Looks good, one last tiny thing

PR #507, opened 2025-04-07
feat: add general access enum to consts

tobice:
We plan to allow users to configure through API & clients how their storages and runs can be accessed — whether just the ID / name is sufficient to access the storage. Full context in https://github.com/apify/apify-core/issues/19012.

This PR adds appropriate enums that will be used both externally (in clients & API), and internally (where it will replace the [existing](https://github.com/apify/apify-core/blob/ba052eedd7490952fa1b8c1447c261a9cfd3acb3/src/packages/consts/src/iam.ts#L101) enum).

Note that there are separate enums for runs and storages as only storages support access by name. So far we have been using a shared enum for both, but it's cleaner to have them separated — in run related logic & documentation we won't have to account (much) for the unsupported value.

### DIFF
--- a/packages/consts/src/consts.ts
+++ b/packages/consts/src/consts.ts
@@ -1,4 +1,4 @@
-import { ValueOf } from './helpers';
+import type { ValueOf } from './helpers';
 import { DNS_SAFE_NAME_REGEX, EMAIL_REGEX } from './regexs';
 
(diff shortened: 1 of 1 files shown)


B4nan (approved):

danpoletaev (approved):

B4nan:
@tobice I've merged the latest changes from master which are updating eslint config, there is one small issue to be fixed here.

PR #508, opened 2025-04-07
chore: update eslint to v9

### LINKED ISSUE #489: Update eslint to v9, use new flat Apify eslint config
We should update eslint to v9 along with the new flat Apify eslint config that supports it.

See [notion.so/apify/New-linting-and-formatting-configuration-across-Apify-0b62890892cd4b71b4fe2dee9291b4af?pvs=4#6152e086487441f6b974e6fbda9614be](https://www.notion.so/apify/New-linting-and-formatting-configuration-across-Apify-0b62890892cd4b71b4fe2dee9291b4af?pvs=4#6152e086487441f6b974e6fbda9614be) for a detailed guide.

B4nan:
Disabled a few of the rules that would be too time-consuming to fix (with dozens of errors for each), otherwise the changes are mostly from the fixer, plus a few small final touches.

Also updated TS, which required one manual adjustment (and will unlock the lockfile update PR from renovate).

Closes #489


jirimoravcik (approved):
Thank you!

PR #509, opened 2025-04-08
feat: fileUrl as string editor type

valekjo:
We want to add file upload to the input schema, decided to go with url on the input.

Seems like a good idea to just use a special editor for this, instead of adding a new input type. The default fallback editor should work reasonably well everywhere.

### Related PRs:
 - https://github.com/apify/apify-docs/pull/1524

### DIFF
--- a/packages/input_schema/src/schema.json
+++ b/packages/input_schema/src/schema.json
@@ -88,7 +88,7 @@
                 "title": { "type": "string" },
                 "description": { "type": "string" },
                 "nullable": { "type": "boolean" },
-                "editor": { "enum": ["javascript", "python", "textfield", "textarea", "datepicker", "hidden"] },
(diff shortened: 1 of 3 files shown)


gippy:
Maybe please also create a proposal for this as a PR to docs and tag Honza Curn and Mara there. I want to avoid the same situation we had with datepicker. I think a file URL which is a string by default with no validation and our UI handling it with upload to KV store makes sense and should be safe and fine with Honza and Mara too.

jancurn:
Good idea, having just an editor with file uploader if much better than extending the schema with some new stuff. Just a note - let's not call the editor `fileUrl` because that makes it look like it's a URL. IMO better name is `fileUpload` of `fileUploader`, and for consistency with the other fields, it should probably be `fileupload` unless those texts are case insensitive

jancurn (approved):

gippy (approved):

PR #514, opened 2025-06-02
feat(schema): add support for resource permissions

tobice:
As part of the [Actor permissions project](https://www.notion.so/apify/Public-Actor-permissions-design-document-1d1f39950a228015a679f276104123cb), we want to allow Actors to request access to resources via schema.

- The Actor input schema already supports annotating string fields with a `resourceType` to specify that the field should reference an Apify platform resource (practically, a storage, such as a dataset).
- This PR extends the input schema so that the Actor can not only specify that a certain string is actually a storage ID, but also what kind of access to the storage the Actor will require (read, or also write). For that reason, the PR adds a new property `resourcePermissions`.
- We will then use this information both to communicate this Actor requirement to the user, and to correctly configure the access when running the Actor.

Example input schema configuration:

```json
{
    "title": "Dataset",
    "type": "string",
    "description": "Select a dataset that you want to process",
    "resourceType": "dataset",
    "resourcePermissions": ["READ", "WRITE"],
}
```

The `resourcePermissions` property follows the same permission format as we use elsewhere in Apify. The Actor developer will be able to specify:

- Nothing → No access at all (works only for full permission Actors).
- `["READ"]`  → The Actor will be able to read the storage.
- `["READ", "WRITE"]` → The Actor will be able to write the storage as well.

We don't support just `["WRITE"]`, which is enforced via schema validation. Later we might choose to introduce "append only" storages.

To address a potential security loophole, the validation also forbids the use of `default` and `prefill` if `resourcePermissions` is set. This is not a breaking change as existing schemas will work just fine until the developer decides to provide `resourcePermissions`.

For additional context refer to the [design doc](https://www.notion.so/apify/Public-Actor-permissions-design-document-1d1f39950a228015a679f276104123cb).

### DIFF
--- a/test/input_schema.test.ts
+++ b/test/input_schema.test.ts
@@ -287,6 +287,25 @@ describe('input_schema.json', () => {
         });
 
         describe('special cases for resourceProperty', () => {
+            it('should accept valid resourceType', () => {
+                const schema = {
+                    title: 'Test input schema',
+                    type: 'object',
+                    schemaVersion: 1,
+                    properties: {
+                        myField: {
+                            title: 'Field title',
+                            description: 'My test field',
+                            type: 'string',
+                            resourceType: 'keyValueStore',
+                            prefill: 'test',
+                            default: 'test',
+                        },
+                    },
+                };
+                expect(() => validateInputSchema(validator, schema)).not.toThrow();
+            });
+
             it('should not accept invalid resourceType', () => {
                 const schema = {
                     title: 'Test input schema',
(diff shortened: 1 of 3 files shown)


tobice:
This is an important change, hence including quite a few folks as reviewers.

test/input_schema.test.ts
```diff
@@ -326,6 +362,181 @@ describe('input_schema.json', () => {
...
+                            type: 'string',
+                            resourceType: 'keyValueStore',
+                            resourcePermissions: ['READ'],
+                            prefill: 'some-value',
+                        },
+                    },
+                };
+
+                expect(() => validateInputSchema(validator, schema)).toThrow(
+                    'Input schema is not valid (Field schema.properties.myField. must NOT be valid)',
```
fnesveda, line 476: That's an ugly error, can we make it prettier somehow?

fnesveda (changes requested):
Looks good, except for the error messages about invalid `default` and `prefill`, can we improve those somehow?

stepskop (approved):
Solid test coverage, looks good 🍻

test/input_schema.test.ts
fnesveda: 😄 AJV is great. OK, let's leave it like it is 🙂

fnesveda (approved):

MFori (approved):

jancurn (approved):

PR #515, opened 2025-06-05
feat(input_schema): Enable secret objects

MFori:
This is proposal to enable `isSecret` for objects in input. Currently this is available only for `string` fields with editors either `textfield` or `textarea`.
Based on the issue [requirements](https://github.com/apify/apify-core/issues/21369)
> An ideal solution would be if we could encrypt objects

In addition to change of Input Schema's JSON schema change it's required to change also the encryption logic.
If input is `string`, there is no change and it's stored in input in form as `TODO`.
In case of `object` input, we need to keep the value as `object` because of validation of input. This propose to store encrypted objects as object with structure:
```
{
  "secret": "encrypted-json-string-of-original-object"
}
```

### DIFF
--- a/packages/input_schema/src/schema.json
+++ b/packages/input_schema/src/schema.json
@@ -237,25 +237,58 @@
         "objectProperty": {
             "title": "Object property",
             "type": "object",
-            "additionalProperties": false,
+            "additionalProperties": true,
             "properties": {
                 "type": { "enum": ["object"] },
                 "title": { "type": "string" },
                 "description": { "type": "string" },
-                "default": { "type": "object" },
-                "prefill": { "type": "object" },
-                "example": { "type": "object" },
-                "patternKey": { "type": "string" },
-                "patternValue": { "type": "string" },
-                "nullable": { "type": "boolean" },
-                "minProperties": { "type": "integer" },
-                "maxProperties": { "type": "integer" },
-
                 "editor": { "enum": ["json", "proxy", "hidden"] },
-                "sectionCaption": { "type": "string" },
-                "sectionDescription": { "type": "string" }
+                "isSecret": { "type": "boolean" }
             },
-            "required": ["type", "title", "description", "editor"]
+            "required": ["type", "title", "description", "editor"],
+            "if": {
+                "properties": {
+                    "isSecret": {
+                        "not": {
+                            "const": true
+                        }
+                    }
+                }
+            },
+            "then": {
+                "additionalProperties": false,
+                "properties": {
+                    "type": { "enum": ["object"] },
+                    "title": { "type": "string" },
+                    "description": { "type": "string" },
+                    "default": { "type": "object" },
+                    "prefill": { "type": "object" },
+                    "example": { "type": "object" },
+                    "patternKey": { "type": "string" },
+                    "patternValue": { "type": "string" },
+                    "nullable": { "type": "boolean" },
+                    "minProperties": { "type": "integer" },
+                    "maxProperties": { "type": "integer" },
+                    "editor": { "enum": ["json", "proxy", "hidden"] },
+                    "sectionCaption": { "type": "string" },
+                    "sectionDescription": { "type": "string" },
+                    "isSecret": { "enum": [false] }
+                }
+            },
+            "else": {
+                "additionalProperties": false,
+                "properties": {
+                    "type": { "enum": ["object"] },
+                    "title": { "type": "string" },
+                    "description": { "type": "string" },
+                    "example": { "type": "object" },
+                    "nullable": { "type": "boolean" },
+                    "editor": { "enum": ["json", "hidden"] },
+                    "sectionCaption": { "type": "string" },
+                    "sectionDescription": { "type": "string" },
+                    "isSecret": { "enum": [true] }
+                }
+            }
         },
         "integerProperty": {
             "title": "Integer property",
--- a/packages/input_secrets/src/input_secrets.ts
+++ b/packages/input_secrets/src/input_secrets.ts
@@ -23,13 +23,29 @@ export function getInputSchemaSecretFieldKeys(inputSchema: any): string[] {
 
 /**
  * Encrypts input secret value
+ * Depending on the type of value, it returns either a string (for strings) or an object (for objects) with the `secret` key.
  */
-export function encryptInputSecretValue({ value, publicKey }: { value: string, publicKey: KeyObject }): string {
-    ow(value, ow.string);
+export function encryptInputSecretValue<T extends string | object>({ value, publicKey }: { value: T, publicKey: KeyObject }):
+    T extends string ? string : { secret: string } {
+    ow(value, ow.any(ow.string, ow.object));
     ow(publicKey, ow.object.instanceOf(KeyObject));
 
-    const { encryptedValue, encryptedPassword } = publicEncrypt({ value, publicKey });
-    return `${ENCRYPTED_INPUT_VALUE_PREFIX}:${encryptedPassword}:${encryptedValue}`;
+    type ResultType = T extends string ? string : { secret: string };
+
+    if (typeof value === 'string') {
+        const { encryptedValue, encryptedPassword } = publicEncrypt({ value, publicKey });
+        return `${ENCRYPTED_INPUT_VALUE_PREFIX}:${encryptedPassword}:${encryptedValue}` as ResultType;
+    }
+
+    let valueStr: string;
+    try {
+        valueStr = JSON.stringify(value);
+    } catch (err) {
+        throw new Error(`The input value could not be stringified for encryption: ${err}`);
+    }
+    // For objects, we return an object with the encrypted JSON string under the 'secret' key.
+    const encryptedJSONString = encryptInputSecretValue({ value: valueStr, publicKey });
+    return { secret: encryptedJSONString } as ResultType;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


gippy:
Ah sorry, the first one is my fault, I forgot that cookies are an array when specifying it

valekjo:
Hey,
sorry, I won't be able to properly review it. I don't have a super strong opinion on this, but I'm more inclined to encoding the values in a different way - here is why :)

I understand that the representation was chosen so that the basic type is always matching, and it's not a problem for the validation.

But the basic type (string/array/object) is all that can be validated, nothing else - so it's not very strong.

So instead I'd go with something like `ENCRYPTED_JSON_VALUE:...`, which would be always string - because it's passed around as string. The user can input anything (meaning the editor can be whatever), and the Actor would decrypt and parse those fields.

Until it's decrypted, nobody should know what's inside (except the user).

From the perspective of types, you'd have `string` on the input, and `unknown` when you parse it in the Actor (it would be up to the author to deal with that).

We can also use that later to cover the string case we have now.

To compare it, the solution in the PR would look like this when you pass it encrypted:

```
{
	stringField: encrypted,
	objectField: { encrypted },
	arrayField: [ encrypted ],
}	
```

And like this when you pass it un-encrypted:

```
{
	stringField: raw,
	objectField: raw,
	arrayField: raw
}
```

Meaning that the encrypted case looks weird - we need a special representation for every type.



I'd be more inclined to

```
{
	stringField: encrypted,
	objectField: encrypted,
	arrayField: encrypted,
}
```

And if passed as not-encrypted:

```
{
	stringField: JSON.stringify(raw),
	objectField: JSON.stringify(raw),
	arrayField: JSON.stringify(raw),
}
```

Meaning that the not-encrypted case looks weird - which is fine, because the encrypted is the default.


I won't make it to the rest of the discussion, just wanted to chip in with this :)

MFori:
> So instead I'd go with something like `ENCRYPTED_JSON_VALUE:...`, which would be always string - because it's passed around as string. The user can input anything (meaning the editor can be whatever), and the Actor would decrypt and parse those fields.

I don't see any additional value having the `ENCRYPTED_JSON_VALUE:...` prefix in the encrypted string over the `ENCRYPTED_VALUE` because we wouldn't parse it anywhere as:
- on FE the secret value is write only, we never open the editor and fill the stored value in
- in Actor's `Actor.getInput()` we still need to return `string` representation, because as a developer I would expect the value match the type of the field - so again no parsing JSON string to object (it's up to Actor developer to parse it).

Generally speaking with you suggestion we still wouldn't support encryption objects and arrays. 

Only one thing we can do is to add support for `json` editor for secret string input, which would behave in UI as an object input. But running via API would require to stringify the input.

jancurn:
I only read this quickly, but I feel:

```
{
	stringField: "secret_value:...",
	objectField: {  "secret": "secret_value:..."" },
	arrayField: [ {  "secret": "secret_value:..."" } ],
}	
```

is a weird "Potemkin" JSON, which will not even be compatible with objects produced by our editors and now the newly introduced sub-schemas, so it only solves some cases, but not the others.

> The issue is that if the type of the field is defined as object then JSON Schema (and Ajv) validates the value to be object and client (Actor) expect to get object from Actor.getInput(). This is fulfilled if the value is not encrypted, but once we want to encrypt it and do just JSON.stringify (and encrypt this string value) the value would be string and JSON schema validation would be failing.

Is this really a problem? When you call an Actor via API, you pass a full unencrypted JSON input, which is validated as is, and encrypted when storing to KV-store. It's only when Actor one would fetch the JSON from KV-store they will receive an object which will not pass validation. But everyone should use `Actor.getValue()` which auto-decrypts it and returns valid object also, no?

MFori:
> Is this really a problem? When you call an Actor via API, you pass a full unencrypted JSON input, which is validated as is, and encrypted when storing to KV-store. It's only when Actor one would fetch the JSON from KV-store they will receive an object which will not pass validation. But everyone should use `Actor.getValue()` which auto-decrypts it and returns valid object also, no?

Yes when you're calling an Actor via API, this should work as you are describing, but when you would start the Actor from UI, the encrypted input is stored right as you pasted it to UI input and this encrypted value is then used to call the API. And in this step we would be sending `string` where `object` is expected. I did this diagram to showcase the issue:

secret_input

What we can do is to catch the error produced by JSON Schema and check if the field has `isSecret` set to `true` and if the string looks like encrypted object. Here we can use the @valekjo's suggestion with `ENCRYPTED_JSON_VALUE` prefix. I'm not sure how easy would be to catch the error from Ajv - it's probably doable, but then the input wouldn't be "valid" against the schema without our additional logic.


> which will not even be compatible with objects produced by our editors and now the newly introduced sub-schemas, so it only solves some cases, but not the others.

I supposed that secret field wouldn't be able to define sub-schema, same as we're doing with properties like `pattern`, `minLength` and `maxLength` which are forbidden once the string input is marked as secret. Object properties like `patternKey`, `patternValue`, `minProperties` and `maxProperties` are already excluded from secret object property in this PR.

jancurn:
Thanks for the clarification. But it really feels like we're changing the external logic (and consequently the Actor specification that we want to make an open standard) just to make some internal implementation easier for us, which is a short-sighted way to design open systems.

I'm sure there's some smarter way to do it. For example, we could prefix the encrypted JSON with `ENCRYPTED_JSON/xyz:val`, where `xyz` would be a hash of the field's schema (skipping non-important properties like `description` and ordering keys alphabetically for stable hashes). This way, we could validate that the schema changed and the encrypted value might no longer match, indicating it somehow in UI to the users.

Also I don't think it's a good idea to block sub-schemas for encrypted objects or arrays. For example, we want to use this new functionality for storing cookies, which need to have a specific format. If we don't validate what users enter in UI or use in API, it will be just another source of user errors and bad UX. We probably cut this corner back in the days when we designed encrypted value for strings, but let's not perpetuate that hack in the future if we can do it better now.

CCing @mtrunkat for visibility

mtrunkat:
I gave it a few more thoughts. I don't like the confusing idea of:

```
{
	stringField: "secret_value:...",
	objectField: {  "secret": "secret_value:..."" },
	arrayField: [ {  "secret": "secret_value:..."" } ],
}	
```

Which is there just to allow us to use an unmodified JSONSchema validator and save some work. I am neither a big fan of having `type: 'object' and passing a prefixed string there. Because then you can't trust the schema when, for example, pushing all the inputs to MongoDB. Those object fields do not contain an object.

I see two more options for this problem -

[1] What about doing it the other way around. Encrypted object will always be `string` that's given and correct. So for cookies, we would use the following:

```
        "myobject": {
            "title": "some object",
            "type": "string", 
            "editor": "json",
            "parsableAs": "object",
            "isSecret": true
        },

        "cookies": {
            "title": "cookies",
            "type": "string",
            "editor": "json",
            "parsableAs": "array",
            "isSecret": true
        },
```

`editor: 'json'` ensures that the value is parsable as JSON and `parsableAs` would make sure the submitted value as JSON-parsable as `parsableAs`. The other option would be to enable developers to define a JSON schema for this object.

At the SDK level, we could automatically parse the value when it has this property.

[2] The other option is field-level encryption. This might be complex for largely nested objects, but it would preserve the shape. https://www.npmjs.com/package/prisma-field-encryption

jancurn:
Seems we're in a bit of a deadlock here...

> [1] What about doing it the other way around. Encrypted object will always be string that's given and correct. So for cookies, we would use the following:

Imagine how this would look like in SDK or API examples ... people would think it's a broken design and it would be a source of errors when callers forget to format the input. Also LLMs would struggle with this, as it's non-standard.

> [2] The other option is field-level encryption. This might be complex for largely nested objects, but it would preserve the shape. https://www.npmjs.com/package/prisma-field-encryption

I agree with @MFori, this leaks too much information and doesn't seem very secure

I don't think it's a problem that encrypted value will no longer conform to some schema. It's pretty expected actually - any entity that is encrypted loses a structure and turns into a raw byte array, and we just do the same. So I'd go with the encrypt to string and capture schema hash approach...

mtrunkat:
Fair, let's move forward this way.

packages/input_schema/src/schema.json
```diff
@@ -153,7 +153,10 @@
                     "title": { "type": "string" },
                     "description": { "type": "string" },
                     "example": { "type": "string" },
+                    "pattern": { "type": "string" },
                     "nullable": { "type": "boolean" },
+                    "minLength": { "type": "integer" },
+                    "maxLength": { "type": "integer" },
```
MFori, line 159: We can now validate these for `secret` strings too.

jancurn (commented):
Look good, just few small comments

packages/input_secrets/src/field_schema_utils.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Generates a stable hash for the field schema.
+ * @param fieldSchema
+ */
+export function getFieldSchemaHash(fieldSchema: Record<string, any>): string {
+    try {
+        const stringifiedSchema = JSON.stringify(normalizeFieldSchema(fieldSchema));
+        // Create a SHA-256 hash of the stringified schema and return the first 10 characters in base64.
+        return crypto.createHash('sha256').update(stringifiedSchema).digest('base64').slice(0, 10);
```
jancurn, line 39: Pls let's change `base64` to `hex`, it looks bit better

valekjo (approved):
Looks good 👍

packages/input_secrets/src/field_schema_utils.ts
```diff
@@ -0,0 +1,42 @@
+import crypto from 'node:crypto';
+
+/**
+ * These keys are omitted from the field schema normalization process
+ * because they are not relevant for validation of values against the schema.
+ */
+const OMIT_KEYS = new Set(['title', 'description', 'sectionCaption', 'sectionDescription', 'nullable', 'example', 'editor']);
```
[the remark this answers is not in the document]
fnesveda: `prefill` is still missing from the `OMIT_KEYS`, and I believe `default` should also be there.

packages/input_schema/src/utilities.ts
```diff
@@ -249,6 +271,16 @@ export function validateInputUsingValidator(
...
         }
 
+        // Additional validation for secret fields
+        if (isSecret && value && typeof value === 'string') {
+            // If the value is a valid encrypted string for the field type,
+            // we check if the field schema is likely to be still valid (is unchanged from the time of encryption).
+            if (isEncryptedValueForFieldType(value, type) && !isEncryptedValueForFieldSchema(value, properties[property])) {
+                // If not, we add an error message to the field errors and user needs to update the value in the input editor.
+                fieldErrors.push(m('inputSchema.validation.secretFieldSchemaChanged', { fieldKey: property }));
+            }
```
fnesveda, line 281: I have a slight concern here, but I'm not sure how big of a problem it will really be.

If you have e.g. an `array` secret field, and you increase its `maxItems`, the encrypted field value will still be semantically valid, but this validation will start throwing errors, which will be annoying. You'll have to hunt for all the passwords / secret keys you put in the field before, some of which might not be available anymore (since some secrets are only shown once after generation).

I'm not sure what changes to the input schema happen more often, but I would wager it's backwards compatible ones, so we'll be forcing people to update their inputs even when not necessary.

fnesveda (changes requested):
Looks good, just two small comments (one of them is a concern rather than a thing to surely fix).

packages/input_schema/src/utilities.ts
MFori: Hm, maybe showing this only as a warning in the UI would be better? 

Ajv does not have anything like warning, but we can updated the `validateInputUsingValidator` function to return list of errors and list of warnings and show this warnings in the UI.

What do you think?

PR #519, opened 2025-06-11
feat(input_schema): Enable sub-schemas in input-schema

mfori:
# Input sub-schemas
> ~~**Note**: This is a proposal, not a final implementation intended for immediate merging in this state. The purpose of this PR is to present a suggested solution for discussion and approval, not the completed implementation.~~

The PR is ready to be merged, few changes from the original proposal were made based on the discussion here:
- validation if the sub-schema is compatible with the editor (e.g. `keyValue` editor should define object with two string properties `key` and `value` and nothing else)
- `schemaBased` editor can be used only in root properties

## 🎯  Goal
The goal of this proposal is to enable creators to define sub-schemas within an Actor's input schema for fields of type `array` and `object`. These field types would support specifying their inner structure, which would be used both for input validation and for rendering the corresponding (sub)fields in the input UI form.

## 📝  Solution
The proposed solution leverages "native" features of JSON Schema, utilizing its properties such as `properties`, `required`, `additionalProperties` (for object) and `items` (for array).

As a result of this, creators would be able to define input schema like this:
```
{
  "title": "Apify Actor input schema example",
  "type": "object",
  "schemaVersion": 1,
  "properties": {
    "my-object": {
      "type": "object",
      "title": "My object",
      "editor": "schemaBased",
      "properties": {
        "key1": {
          "type": "string",
          "title": "Key 1",
          "description": "Description",
          "editor": "textfield"
        },
        "key2": {
          "type": "integer",
          "title": "Key 2",
          "description": "Description",
          "editor": "integer"
        }
      },
      "required": ["key1"],
      "additionalProperties": false
    },
    "my-array": {
      "type": "array",
      "title": "My array",
      "editor": "json",
      "items": {
        "type": "object",
        "properties": {
          "key1": {
            "type": "string",
            "title": "Key 2",
            "description": "Description",
            "editor": "textfield"
          },
          "key2": {
            "type": "integer",
            "title": "Key 2",
            "description": "Description",
            "editor": "integer"
          }
        },
        "required": ["key1"],
        "additionalProperties": false
      }
    }
  },
  "required": []
}
```

Actor with schema like this, would then accept input like:
```
{
  "my-object": {
    "key1": "test",
    "key2": 123
  },
  "my-array": [
    {
      "key1": "test", 
      "key2": 123
    },
    {
      "key1": "test" 
    }
  ]
}
```

### Recursiveness
The schema should support recursion, allowing creators to define nested objects within other objects. This enables complex and deeply structured input definitions as needed.

```
{
  "title": "Apify Actor input schema example",
  "type": "object",
  "schemaVersion": 1,
  "properties": {
    "my-object": {
      "type": "object",
      "title": "My object",
      "editor": "schemaBased",
      "properties": {
        "key1": {
          "type": "string",
          "title": "Key 1",
          "description": "Description",
          "editor": "textfield"
        },
        "key2": {
          "type": "object",
          "title": "Key 2",
          "description": "Description",
          "editor": "schemaBased",
          "properties": {
            "subKey": {
              "type": "string",
              "title": "SubKey",
              "description": "Description",
              "editor": "textfield"
            }
          }
        }
      },
      "required": ["key1"],
      "additionalProperties": false
    },
  "required": []
}
```

The same goes with arrays:
```
{
  "title": "Apify Actor input schema example",
  "type": "object",
  "schemaVersion": 1,
  "properties": {
    "my-array": {
      "type": "array",
      "title": "My array",
      "editor": "schemaBased",
      "items": {
        "type": "array",
        "items": {
          "type": "string"
        }
      }
    }
  }
  "required": []
}
```

### JSON Schema
At the JSON Schema level, the implementation is relatively straightforward and can be basically follow the approach used in this PR. The proposed changes include:
- **Creating new definitions for each property type** - some properties used in the root schema don’t make sense within a sub-schema context. Therefore, instead of reusing the root definitions with complex conditions, it’s simpler to create tailored definitions for sub-schemas.
- **Extending the `object` and `array` definitions with new properties**:
  - `object` type can include:
    - `properties` - defines the internal structure of the object. It supports all property types available at the root input schema level (with mentioned restrictions).
    - `additionalProperties` - specifies whether properties not listed in `properties` are allowed
    - `required` - lists which properties defined in `properties` are required
    
      ```
      {
        "type": "object",
        "properties": {
          "key": {
            ...
          }
        },
        "additionalProperties": false,
        "required": "key"
      }
      ```    

  - `array` type can include
    - `items` - defines the type (and optionally the shape) of array items
      
      ```
      {
        "type": "array",
        "items": {
          "type": "object",
          "properties": {...}
        }
      }
      ```

#### Actor's Input Schema
Validation would almost work out of the box with the updated `schema.json` file, but a few additional steps are required:
- Since we're manually validating all properties one by one against the schema (using the `validateProperties` function), we also need to validate sub-properties. The validation has to be done against a different set of definitions in case of root-level properties and sub-properties. This should be straightforward to implement.
- the logic in `parseAjvError` needs to be updated to correctly display the path to the relevant (sub)property in validation errors. Again, this is not expected to be complex.

#### Input
Because all newly added properties (`properties`, `additionalProperties`, `required` and `items`) are native features of JSON Schema, input validation against the Actor's input schema will work entirely out of the box.

### Input UI
In the Input UI, we want to give creators the flexibility to render each sub-field of a complex input field independently.

A proof of concept has already been implemented (currently only for object fields), and there are no major blockers to a full implementation. You can see the draft PR here: https://github.com/apify/apify-core/pull/21564

> Note: The code in the PR is intentionally minimal, not optimal and not production-ready. Its purpose is to validate the approach.

Creators should have the option to choose how a field is rendered:
- Use an existing editor (e.g. `json`), in which case the sub-schema is used solely for validation.
- Or render each sub-property as an individual input field based on the sub-schema. 

To support the latter, we need to introduce a new editor type that signals sub-schema-based rendering. I’ve tentatively called this editor `schemaBased`, but the name is open for discussion.

For arrays using sub-schemas with the `schemaBased` editor, we’ll need a UI component that includes controls to add, remove, and optionally reorder items.

**Note**: Based on the discussion below, we decide to limit `schemaBased` editor only for root level properties.

**Technical Implementation Notes**
- The main change in the Input UI will be to recursively render sub-fields when the `schemaBased` editor is used.
- We’ll use dot notation (e.g. field.subField) for Formik field names to ensure proper binding. Formik handles this automatically.
- We'll also need to support labels, descriptions, and other for sub-fields, but this should be relatively straightforward.

### Root-level properties
We are effectively reusing the existing "root-level" property definitions within sub-schemas. However, not all root-level properties make sense in context of sub-schema. Specifically, the following properties are questionable:
- `default`, `prefill` and `example` - These are better suited for the root property that contains the entire object or array. Applying them to individual sub-fields could lead to unexpected or inconsistent behavior.
- `sectionCaption` and `sectionDescription` - These introduce structural elements (sections) and may not make sense inside nested sub-schemas. We should consider either removing them entirely from sub-schemas or revisiting their design from a UI/UX perspective (e.g. nested sections).

### `schemaBased` editor
As mentioned in the Input UI section, there is a need to introduce a new editor type that signals that each sub-property of a complex object or array should be rendered as a standalone field using its sub-schema definition.

I’ve proposed the name `schemaBased` for this new editor, but the name is open for discussion.

### Compatibility between editor and sub-schema
A key question is whether we should validate the compatibility between the `editor` and the defined sub-schema within the Actor's input schema, or leave this responsibility to the Actor developer.

#### Example scenario
A developer defines a property of type `array` with `editor: stringList`, but also provides a sub-schema specifying object-type items. The input UI would generate a list of strings, while the validation would expect objects, resulting in invalid input.

Possible Approaches:
1. **No restrictions (responsibility on creator)**
  Allow any combination of `editor` and sub-schema, and assume the creator understands which combinations are valid. This offers maximum flexibility but increases the risk of misconfiguration.
2. **Restrict available editors when a sub-schema is defined** 
  that would be `schemaBased`, `json` and `hidden`.
3. **Strict validation based on editor type**
  Enforce that sub-schemas match expected structures for specific editors. For example for `stringList` editor the sub-schema can only be string typed items, for `requestListSources` it's object with strictly defined properties. But this would make the JSON Schema way more complicated with lot's of if-else branches and duplicated properties definitions.

> Note to this: We are currently validating structure of input for some editors (for example `requestListSources`) manually in `validateInputUsingValidator`. So in this case `editor` is not used just for UI but also influence the validation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,16 +63,15 @@
             "link": true
         },
         "node_modules/@apify/eslint-config": {
-            "version": "1.0.1",
-            "resolved": "https://registry.npmjs.org/@apify/eslint-config/-/eslint-config-1.0.1.tgz",
-            "integrity": "sha512-pYTAaWYyvcY1h0K871bFEU94h2nbLkkQMHmRWOc3VmMaKNAqPLHPclEi+mx+YL+yYoCI/YHYeg3P3yGT3sp/Ww==",
+            "version": "1.1.0",
+            "resolved": "https://registry.npmjs.org/@apify/eslint-config/-/eslint-config-1.1.0.tgz",
+            "integrity": "sha512-3rVBzGcRP13aN8pcYz8qkbJ7ji4oJ/haabaiRiI43ytKBwasMA4rAcH0ZifVKm2qOAOj5Zd0e9YSuiqlTPDoDg==",
             "dev": true,
             "license": "ISC",
             "dependencies": {
                 "@eslint/compat": "^1.2.6",
-                "@jirimoravcik/eslint-plugin-import": "2.32.0",
                 "eslint-config-airbnb-base": "^15.0.0",
-                "eslint-plugin-import": "^2.31.0",
+                "eslint-plugin-import": "^2.32.0",
                 "eslint-plugin-simple-import-sort": "^12.1.1",
                 "globals": "^15.14.0"
             },
@@ -199,22 +198,14 @@
                 "url": "https://opencollective.com/babel"
             }
         },
-        "node_modules/@babel/core/node_modules/debug": {
-            "version": "4.4.1",
-            "resolved": "https://registry.npmjs.org/debug/-/debug-4.4.1.tgz",
-            "integrity": "sha512-KcKCqiftBJcZr++7ykoDIEwSa3XWowTfNPo92BYxjXiyYEVrUQh2aLyhxBCwww+heortUFxEJYcRzosstTEBYQ==",
+        "node_modules/@babel/core/node_modules/semver": {
+            "version": "6.3.1",
+            "resolved": "https://registry.npmjs.org/semver/-/semver-6.3.1.tgz",
+            "integrity": "sha512-BR7VvDCVHO+q2xBEWskxS6DJE1qRnb7DxzUrogb71CWoSficBxYsiAGd+Kl0mmq/MprG9yArRkyrQxTO6XjMzA==",
             "dev": true,
-            "license": "MIT",
-            "dependencies": {
-                "ms": "^2.1.3"
-            },
-            "engines": {
-                "node": ">=6.0"
-            },
-            "peerDependenciesMeta": {
-                "supports-color": {
-                    "optional": true
-                }
+            "license": "ISC",
+            "bin": {
+                "semver": "bin/semver.js"
             }
         },
         "node_modules/@babel/generator": {
@@ -249,6 +240,16 @@
             },
             "engines": {
                 "node": ">=6.9.0"
+            }
+        },
+        "node_modules/@babel/helper-compilation-targets/node_modules/semver": {
+            "version": "6.3.1",
+            "resolved": "https://registry.npmjs.org/semver/-/semver-6.3.1.tgz",
+            "integrity": "sha512-BR7VvDCVHO+q2xBEWskxS6DJE1qRnb7DxzUrogb71CWoSficBxYsiAGd+Kl0mmq/MprG9yArRkyrQxTO6XjMzA==",
+            "dev": true,
+            "license": "ISC",
+            "bin": {
+                "semver": "bin/semver.js"
             }
         },
         "node_modules/@babel/helper-globals": {
@@ -636,28 +637,10 @@
                 "node": ">=6.9.0"
             }
         },
-        "node_modules/@babel/traverse/node_modules/debug": {
-            "version": "4.4.1",
-            "resolved": "https://registry.npmjs.org/debug/-/debug-4.4.1.tgz",
-            "integrity": "sha512-KcKCqiftBJcZr++7ykoDIEwSa3XWowTfNPo92BYxjXiyYEVrUQh2aLyhxBCwww+heortUFxEJYcRzosstTEBYQ==",
-            "dev": true,
-            "license": "MIT",
-            "dependencies": {
-                "ms": "^2.1.3"
-            },
-            "engines": {
-                "node": ">=6.0"
-            },
-            "peerDependenciesMeta": {
-                "supports-color": {
-                    "optional": true
-                }
-            }
-        },
         "node_modules/@babel/types": {
-            "version": "7.28.0",
-            "resolved": "https://registry.npmjs.org/@babel/types/-/types-7.28.0.tgz",
-            "integrity": "sha512-jYnje+JyZG5YThjHiF28oT4SIZLnYOcSBb6+SDaFIyzDVSkXQmQQYclJ2R+YxcdmK0AX6x1E5OQNtuh3jHDrUg==",
+            "version": "7.28.1",
+            "resolved": "https://registry.npmjs.org/@babel/types/-/types-7.28.1.tgz",
+            "integrity": "sha512-x0LvFTekgSX+83TI28Y9wYPUfzrnl2aT5+5QLnO6v7mSJYtEEevuDRN0F0uSHRk1G1IWZC43o00Y0xDDrpBGPQ==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -779,19 +762,6 @@
             },
             "engines": {
                 "node": ">=v18"
-            }
-        },
-        "node_modules/@commitlint/is-ignored/node_modules/semver": {
-            "version": "7.7.2",
-            "resolved": "https://registry.npmjs.org/semver/-/semver-7.7.2.tgz",
-            "integrity": "sha512-RF0Fw+rO5AMf9MAyaRXI4AV0Ulj5lMHqVxxdSgiVbixSCXoEmmX/jk0CuJw4+3SqroYO9VoUh+HcuJivvtJemA==",
-            "dev": true,
-            "license": "ISC",
-            "bin": {
-                "semver": "bin/semver.js"
-            },
-            "engines": {
-                "node": ">=10"
             }
         },
         "node_modules/@commitlint/lint": {
@@ -1517,24 +1487,6 @@
                 "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
             }
         },
-        "node_modules/@eslint/config-array/node_modules/debug": {
-            "version": "4.4.1",
-            "resolved": "https://registry.npmjs.org/debug/-/debug-4.4.1.tgz",
-            "integrity": "sha512-KcKCqiftBJcZr++7ykoDIEwSa3XWowTfNPo92BYxjXiyYEVrUQh2aLyhxBCwww+heortUFxEJYcRzosstTEBYQ==",
-            "dev": true,
-            "license": "MIT",
-            "dependencies": {
-                "ms": "^2.1.3"
-            },
-            "engines": {
-                "node": ">=6.0"
-            },
-            "peerDependenciesMeta": {
-                "supports-color": {
-                    "optional": true
-                }
-            }
-        },
         "node_modules/@eslint/config-helpers": {
             "version": "0.3.0",
             "resolved": "https://registry.npmjs.org/@eslint/config-helpers/-/config-helpers-0.3.0.tgz",
@@ -1546,9 +1498,9 @@
             }
         },
         "node_modules/@eslint/core": {
-            "version": "0.14.0",
-            "resolved": "https://registry.npmjs.org/@eslint/core/-/core-0.14.0.tgz",
-            "integrity": "sha512-qIbV0/JZr7iSDjqAc60IqbLdsj9GDt16xQtWD+B78d/HAlvysGdZZ6rpJHGAc2T0FQx1X6thsSPdnoiGKdNtdg==",
+            "version": "0.15.1",
+            "resolved": "https://registry.npmjs.org/@eslint/core/-/core-0.15.1.tgz",
+            "integrity": "sha512-bkOp+iumZCCbt1K1CmWf0R9pM5yKpDv+ZXtvSyQpudrI9kuFLp+bM2WOPXImuD/ceQuaa8f5pj93Y7zyECIGNA==",
             "dev": true,
             "license": "Apache-2.0",
             "dependencies": {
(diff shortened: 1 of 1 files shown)


gippy (approved):
Good job, I'm not 100% sure about the `schemaBased` editor, maybe `subfields` as editor name would work better?

Also I think we should lock it to only allow one level of recursivity, otherwise the UI might be pretty painful if someone missuses it.

jancurn:
And one more note: now that we have this, I think we should deprecate `patternKey` and `patternValue` and other similar relicts from the past, and get closer to JSON schema which is an industry standard, widely understood by LLMs.

mtrunkat:
@gippy, @MFori, please add me to these proposals next time.

I see you reached some conclusion, but I want to raise one thing:

People who are using our UI are not our core persona. People who are integrating us are so a good question is if this type of input will work for those people?
- Integration via API? I believe they will make it if we invest a lot of time in documenting this well on the Actor's page
- Integration via Zapier? Are we able to render this in Zapier? If not we will just display a textarea there. But the users will likely miss there a guidance. Have we think about this?
- Integration via MCP? I am not sure if Claude will be able to fill in such input, and it would be worth giving it a shot.

Building a UI internally for this is easy, but that is not how people use Apify, and we need to make sure that all the features here will work in all 3 (and more) cases above.

CleanShot 2025-06-17 at 14 19 31@2x

jancurn:
I agree with the above - we already accept `object` and `array` without any schema, so adding a schema will only improve the situation, not make anything worse. The Actor creators still need to understand that more complex objects will make integrations harder.

Just last small change - please let's call the editor `subschema`, for consistency with other editor names.

packages/input_schema/src/input_schema.ts
```diff
@@ -38,6 +38,11 @@ export function parseAjvError(
     let fieldKey: string;
     let message: string;
 
+    // remove leading and trailing slashes and replace remaining slashes with dots
+    const cleanPropertyName = (name: string) => {
+        return name.replace(/^\/|\/$/g, '').replace(/\//g, '.');
+    };
```
MFori, line 44: This is needed, because now error can be related to sub-properties and we want to show nice path.

packages/input_schema/src/input_schema.ts
```diff
@@ -92,11 +100,33 @@ function validateBasicStructure(validator: Ajv, obj: Record<string, unknown>): a
...
+ * @param subField If true, the field is a sub-field of another field, so we need to skip some definitions.
  */
-function validateField(validator: Ajv, fieldSchema: Record<string, unknown>, fieldKey: string): asserts fieldSchema is FieldDefinition {
+function validateField(validator: Ajv, fieldSchema: Record<string, unknown>, fieldKey: string, subField = false): asserts fieldSchema is FieldDefinition {
     const matchingDefinitions = Object
         .values<any>(definitions) // cast as any, as the code in first branch seems to be invalid
         .filter((definition) => {
+            // Because the definitions contains not only the root properties definitions, but also sub-schema definitions
+            // and utility definitions, we need to filter them out and validate only against the appropriate ones.
+            // We do this by checking prefix of the definition title (Utils: or Sub-schema:)
```
MFori, line 114: I haven't found better approach then grouping the definitions by its title, it's not that robust but works well. 
Another approach would be to define `definitions`, `sub_definitions` and `util_definitions` but that's not valid JSON Schema keywords and wouldn't pass `Ajv` strict validation. 
Or put it as sub objects like `definitions.root`, `definitions.sub`, `definitions.utils`, but haven't test it.

But since this is used just here I think it's ok to go with this solution.

packages/input_schema/src/input_schema.ts
```diff
@@ -168,6 +227,10 @@ export function validateExistenceOfRequiredFields(inputSchema: InputSchema) {
  * then checks that all required fields are present and finally checks fully against the whole schema.
  *
  * This way we get the most accurate error message for user.
+ *
+ * @param validator An instance of AJV validator. Important: The JSON Schema that the passed input schema is validated against
+ *  is using features from JSON Schema 2019 draft, so the AJV instance must support it.
```
MFori, line 232: This PR already more than doubled it's length, but without utilising the `unevaluatedProperties` (https://json-schema.org/understanding-json-schema/reference/object#unevaluatedproperties) from draft `2019-09` it would be much more with lot of duplicated definitions (there is space to update existing definitions and replace `additionalProperties` with `unevaluatedProperties` and make the schema smaller and cleaner).

But to support this the `Ajv` validator instance passed to `validateInputSchema` function has to support this version, so we need update the `Ajv` import in all places that are calling this function from:
```
import Ajv from 'ajv'
```
to
```
import Ajv from 'ajv/dist/2019'
```
It should be these places (we should change it together with bumping version of `@apify/input_schema` once this PR is merged):
- `apify/apify-worker` - when validating schema during build
- `apify/apify-cli` - `validate-schema` command
- `apify/apify-core` - just admin input schema playground

Note: the `draft 2019-09` is used only to validate input-schema against our meta JSON Schema, but input-schema itself has only features from `draft 07` so when validating input against input-schema we don't need to change anything.

MFori:
@gippy @fnesveda @valekjo 
The PR is ready for review, few changes from the original proposal were made based on the discussion here:
- validation if the sub-schema is compatible with the editor (e.g. `keyValue` editor should define object with two string properties `key` and `value` and nothing else)
- `schemaBased` editor can be used only in root properties

The plan is to merge this, update worker and cli with new version (and use JSON Schema draft 2019 for input schema validation) and then:
- update UI to support `schemaBased` editor
- update UI to present the sub-schema in input tab on Actor detail
- update docs

packages/input_schema/src/input_schema.ts
valekjo, line 114: Suggestion: How about not making this part of `validateField` and pre-calculating them as `fieldDefinitions` and `subFieldDefinitions` instead?

packages/input_schema/src/input_schema.ts
```diff
@@ -92,11 +100,33 @@ function validateBasicStructure(validator: Ajv, obj: Record<string, unknown>): a
 
 /**
  * Validates particular field against it's schema.
+ * @param validator An instance of AJV validator (must support draft 2019-09).
+ * @param fieldSchema Schema of the field to validate.
+ * @param fieldKey Key of the field in the input schema.
+ * @param subField If true, the field is a sub-field of another field, so we need to skip some definitions.
  */
-function validateField(validator: Ajv, fieldSchema: Record<string, unknown>, fieldKey: string): asserts fieldSchema is FieldDefinition {
+function validateField(validator: Ajv, fieldSchema: Record<string, unknown>, fieldKey: string, subField = false): asserts fieldSchema is FieldDefinition {
```
valekjo, line 108: Nit: I'd rename the parameter to `isSubField`.

packages/input_schema/src/input_schema.ts
```diff
@@ -121,30 +161,49 @@ function validateField(validator: Ajv, fieldSchema: Record<string, unknown>, fie
...
+ */
+function validateSubFields(validator: Ajv, fieldSchema: InputSchemaBaseChecked, fieldKey: string) {
+    Object.entries(fieldSchema.properties).forEach(([subFieldKey, subFieldSchema]) => {
+        // The sub-properties has to be validated first, so we got more relevant error messages.
+        if ((subFieldSchema as any).properties) {
+            // If the field has sub-fields, we need to validate them as well.
+            validateSubFields(validator, subFieldSchema as any as InputSchemaBaseChecked, `${fieldKey}.${subFieldKey}`);
+        }
+        validateField(validator, subFieldSchema, `${fieldKey}.${subFieldKey}`, true);
+    });
```
valekjo, line 192: Question: Don't we need something like this for the array fields, too?

packages/input_schema/src/input_schema.ts
valekjo: Can we maybe add `createAjv` or some other mechanism that would make sure that the correct schema version is used every time, without the need to check every call of the function? Or maybe there is some property on the Ajv instance that we could check?

test/input_schema.test.ts
```diff
@@ -538,5 +540,300 @@ describe('input_schema.json', () => {
...
+                                        key2: {
+                                            type: 'string',
+                                            title: 'Key 2',
+                                            description: 'Key 2 description',
+                                        },
+                                    },
+                                },
+                            },
+                        },
+                    },
```
valekjo, line 578: Question: This defines inputs like
```
{ myField: {
  key: {
    key1: '',
    key2: '',
  }
} }
```

Would this be enough for the test?

```
{ myField: {
  key1: '',
  key2: '',
} }
```

test/utilities.client.test.ts
```diff
@@ -1271,6 +1271,99 @@ describe('utilities.client', () => {
...
+                    { field: { key1: 'value' } },
+                    { field: { key1: 'value', key2: 'value' } },
+                ];
+
+                const results = inputs
+                    .map((input) => validateInputUsingValidator(validator, inputSchema, input))
+                    .filter((errors) => errors.length > 0);
+
+                // There should be 4 invalid inputs
+                expect(results.length).toEqual(4);
```
valekjo, line 1317: Suggestion: Would change to explicit `validInputs` and `invalidInputs` (it's easier to read and makes the test condition stronger)

valekjo (approved):
Ton of work 💪 I've only checked the schema.json briefly (to understand a bit how it's constructed), but mostly relying on tests.

I remember we had some issues with fields with dots in their name, we should check if that could cause any issues here.

test/input_schema.test.ts
MFori: This is testing that the recursion sub-schema of sub-object works too

packages/input_schema/src/input_schema.ts
MFori: Nice idea, done

packages/input_schema/src/input_schema.ts
MFori: I've add check (`ensureAjvSupportsDraft2019`) that would validate it or throw exception. Do you think it's enough?

MFori:
> I remember we had some issues with fields with dots in their name, we should check if that could cause any issues here.

I've added test for it and it seems it doesn't cause any issues here, but we would definitely need to double check it in UI because it might cause issues with our current implementation of input UI with Formik

B4nan:
apparently this broke the latest version of the CLI

image

PR #520, opened 2025-06-12
feat: add MCP server category

mhamas:
@jancurn how exactly would you like the category to be named? Is `MCP_SERVERS: MCP servers` OK? Thanks


jancurn (approved):
Yeah I can't think of anything better, cheers

PR #527, opened 2025-08-04
feat: add createStorageContentSignature

danpoletaev:
This PR moves `createStorageContentSignature()` to shared package.
Before it was used only in JS client, but now we also need this function in JS SDK.


test/storages.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        const signature = createStorageContentSignature({
+            resourceId: message,
+            urlSigningSecretKey: secretKey,
+        });
+
+        const [version, expiresAt, hmac] = Buffer.from(signature, 'base64url').toString('utf8').split('.');
+        expect(signature).toBe('MC4wLjNUd2ZFRTY1OXVmU05zbVM0N2xS');
+        expect(version).toBe('0');
+        expect(expiresAt).toBe('0');
+        // This HMAC should match HMAC generated by Python function.
```
janbuchar, line 17: From the perspective of a maintainer, it is strange to mention the Python counterpart here. Please either remove the comment or describe the code in a way that works without reading a repository in a different language.

janbuchar (approved):

PR #538, opened 2025-09-18
feat: @apify/json_schemas package

MFori:
### `@apify/json_schemas` package

This introduce new `@apify/json_schemas` package that would hold all our actor schemas definitions, it moves there:
- `input.json` from `@apify/input_schema`
- `actor.json` from worker
- `dataset.json` from worker
- `key_value_store.json` from worker
- `output.json` from worker

This allows us:
- to use the schemas in other places too (e.g. API), not just in worker
- publish them since this repository is public, which will allow validation and other features in IDEs

How to use this package can be seen in this draft PR in worker, which removes the schemas from there and uses this new package: https://github.com/apify/apify-worker/pull/1508

Schema can be imported right away as object:
```
import { actorSchema, inputSchema } from '@apify/json_schemas';
```
Or the `ajv` validator builder function:
```
import { getActorSchemaValidator } from '@apify/json_schemas';
const validator = getActorSchemaValidator();
```

All the schemas are `.json` files except the `actor.ts` and `actor.json`. In this case the `.json` file is generated from the `actor.ts` during the build step. We had it like this already in worker, where `ACTOR_LIMITS` are imported to the schema from `@apify/consts`. But we need public `.json` file too. But I'm not sure how to prevent devs modifying the `.json` file directly 🤔


jest.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
         '<rootDir>/packages/*/src/**/*.ts',
     ],
     moduleNameMapper: {
+        '^@apify/json_schemas/(.*)$': '<rootDir>/packages/json_schemas/$1',
```
B4nan, line 15: why is this needed? looks like the general regex on the next line should cover it just fine

MFori: This is still draft I need to check if everything would work correctly.

But to this jest config - it wouldn't work, because the general regex would convert import like:

```
import schema from '@apify/json_schemas/schemas/input.json';
```

to:

```
/packages/json_schemas/schemas/input.json/src;
```

I would probably export the schemas form the package as constants so it wouldn't be necessary, but later.

B4nan: Deep imports are rather an antipattern, I wouldn't design our packages to depend on that. Why not `import { input } from '@apify/json_schemas'` instead?

B4nan (approved):
didn't read this in detail, I was mostly looking for "meta issues"

packages/json_schemas/schemas/actor.json
tobice, line 1: I wonder, would it make sense to include _schema_  in the schema file names? For example `actor.schema.json`.  

The file might be easier to recognize when searched for...

And in general I have a problem that `actor.json` might be both an Actor definition file, and the schema for that file 😄

tobice (approved):
I assume this mostly just copy-paste / shuffle around, so I didn't go through the actual file contents much.

The change makes sense, thanks for the nice PR description 👏

valekjo (approved):

valekjo:
Looks good! It will be nice to have the schemas exposed publicly.

packages/json_schemas/package.json
```diff
@@ -0,0 +1,52 @@
...
+    },
+    "bugs": {
+        "url": "https://github.com/apify/apify-shared-js/issues"
+    },
+    "homepage": "https://apify.com",
+    "scripts": {
+        "build": "npm run clean && npm run build-schemas && npm run compile && npm run copy",
+        "clean": "rimraf ./dist",
+        "compile": "tsup",
+        "build-schemas": "ts-node scripts/build-schemas.ts",
```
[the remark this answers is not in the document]
B4nan: We could most likely just run TS scripts via node these days. `tsup` is here because we publish dual CJS/ESM builds.

PR #539, opened 2025-09-19
ci: ensure JSON schemas are in sync

mfori:
follow-up to https://github.com/apify/apify-shared-js/pull/538

We need to ensure that for schemas that have `.ts` file used to generate `.json` file:
- users modify the `.ts` file not the `.json` file.
- users run `npm run build` to generate the `.json` file and commits this generated file.
- that means -> `.ts` and `.json` files are always in sync

This adds CI workflow that runs whenever this `.ts` or `.json` file changed, runs the build and check if `.json` file changes or not.


.github/workflows/json_schemas_pr_check.yaml
```diff
@@ -0,0 +1,58 @@
...
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v5
+              with:
+                node-version: ${{ matrix.node-version }}
+
+            - name: Cache node_modules
+              uses: actions/cache@v4
+              with:
+                path: '**/node_modules'
+                key: ${{ runner.os }}-${{ matrix.node-version }}-modules-${{ hashFiles('**/package-lock.json') }}
```
B4nan, line 30: this is likely a copy pasta from other workflow, but here it makes no sense to check it for different node versions (`${{ matrix.node-version }}`)

side note: is that cache action even needed? i believe `setup-node` has native support for caching (and with npm it should actually work fine, unlike with other package managers)

.github/workflows/json_schemas_pr_check.yaml
```diff
@@ -0,0 +1,58 @@
...
+                    if ! git diff --exit-code "$json_file" >/dev/null; then
+                        echo "ERROR: $json_file is out of sync with $ts_file."
+                        echo "Please put the schema changes to .ts file ($ts_file) and update the .json file ($json_file) by running the build script (npm run build)."
+                        failed=1
+                    fi
+                done
+
+                if [ $failed -eq 1 ]; then
+                    exit 1
+                fi
```
B4nan, line 58: This feels a bit overcomplicated, any reason you just don't diff the whole project? It will be fast enough, surely faster than running `npm ci`, building things or checking out the repo itself.

PR #548, opened 2025-10-06
feat(json_schemas): Do not allow colon in field keys in dataset fields schema

MFori:
Do not allow using colon in field keys in dataset fields schema.

There's issue when dataset field in the dataset schemas has colon in the key, because when we're collecting dataset statistics (in `MongoUpdateThrottled`) we put commands to redis in form `{operation}:{mongoDocumentId}:{propertyName}:{value}`, like `inc:{mongoDocumentId}:property:20`. Then when we're actually updating the mongodb document we get these values from redis and split it by `:`. 

So if the `property` has colon inside it breaks.

Reported here: https://apify.slack.com/archives/C0L33UM7Z/p1759569404545289

### DIFF
--- a/packages/json_schemas/schemas/dataset.schema.json
+++ b/packages/json_schemas/schemas/dataset.schema.json
@@ -18,7 +18,20 @@
             "type": "string"
         },
         "fields": {
-            "$ref": "http://json-schema.org/draft-07/schema#"
+            "allOf": [
+                { "$ref": "http://json-schema.org/draft-07/schema#" },
+                {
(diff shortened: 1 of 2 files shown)


fnesveda (changes requested):
Shouldn't we rather escape the `:` characters on the backend, so that we don't have to limit how the dataset items can be shaped? IMHO if JSON objects can contain `:` in the key names, then we should support it. We shouldn't limit our users due to internal implementation complexities.

PR #608, opened 2026-04-01
chore: deprecate `input` and `output` fields in favor of `inputSchema` and `outputSchema`

JJetmar:
Based on [discussion in Slack](https://apify.slack.com/archives/CD0SF6KD4/p1774997458909059) We want to mark `input` and `output` attributes as deprecated. And require mutually exclusive definitions for input/inputSchema and output/outputSchema.

Additionally I also noticed there was `modification-rules-schema.xsd` missing when the logic for describing and bundling json schemas was migrated from previous repository.

### DIFF
--- a/packages/json_schemas/output/actor.ide.json
+++ b/packages/json_schemas/output/actor.ide.json
@@ -122,9 +122,10 @@
                     "$ref": "#/definitions/input-ide-json-2212166aa408babb6120b85f1282521a"
                 }
             ],
-            "description": "Path or direct definition of the Actor's input schema.",
-            "x-intellij-html-description": "<p>Path or direct definition of the Actor's input schema.</p>",
-            "markdownDescription": "Path or direct definition of the Actor's input schema."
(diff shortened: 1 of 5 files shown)


JJetmar:
Actually we may just enchant it with:
```json
{
  ...
  "deprecated": true,
  "description": "Deprecated - please use `input` instead"
}
```

I will check what is the support in the editors first.

JJetmar:
I tested in VSCode and Webstorm. VSCode worked without issues. Webstorm doesn't seems to support `"deprecated": true` statement and the "error" message for mutual exclusion of input/inputSchema and output/outputSchema is not very specific there, but at least it shows the problem:
image

mfori:
Yeah, the same is in monaco editor in console, but I think that from the message box it's clear what the issue is

Screenshot 2026-04-14 at 10 09 46

mfori (approved):
Thanks 👍

PR #609, opened 2026-04-02
feat: add defaultMemoryMbytes to schema

daniil-poletaev:
Add `defaultMemoryMbytes` to schema

### DIFF
--- a/packages/json_schemas/schemas/actor.schema.json
+++ b/packages/json_schemas/schemas/actor.schema.json
@@ -60,6 +60,16 @@
       "minimum": 128,
       "maximum": 32768
     },
+    "defaultMemoryMbytes": {
+      "oneOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "integer"
+        }
+      ]
+    },
     "input": {
       "oneOf": [
         {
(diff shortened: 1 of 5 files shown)


packages/json_schemas/schemas/actor.schema.json
```diff
@@ -60,6 +60,16 @@
       "minimum": 128,
       "maximum": 32768
     },
+    "defaultMemoryMbytes": {
+      "oneOf": [
+        {
+          "type": "string"
```
B4nan, line 66: can it really be a string?

B4nan (approved):
(i dont feel like the right person to review this, but it looks good to me)

packages/json_schemas/schemas/actor.schema.json
daniil-poletaev: Yes, this is actually main type - dynamic memory expression.

packages/json_schemas/src/actor.schema.ts
```diff
@@ -65,6 +65,12 @@ export const actorSchema = {
             minimum: ACTOR_LIMITS.MIN_RUN_MEMORY_MBYTES,
             maximum: ACTOR_LIMITS.MAX_RUN_MEMORY_MBYTES,
         },
+        defaultMemoryMbytes: {
+            oneOf: [
+                { type: 'string' },
+                { type: 'integer' },
```
barjin, line 71: With the `integer` type, why don't we constrain it to the `128 - 32768` limits, as the other fields (`minMemoryMbytes`, etc.)? What about the `enum` with the power-of-2 values?

barjin (commented):
Sorry for a post-merge review, just one curious question:

packages/json_schemas/src/actor.schema.ts
daniil-poletaev: Oh, good point 👍 

Just added a fix to that: https://github.com/apify/apify-shared-js/pull/614

PR #616, opened 2026-04-14
chore: migrate to pnpm and enable minimum release age

B4nan:
## Summary

Pilot of the org-wide supply-chain hardening + pnpm migration plan (first of ~14 repos). This PR migrates `apify-shared-js` from npm workspaces to pnpm workspaces and adds a 1-day minimum release age guard at two layers so newly-published package versions cannot be installed for 24h after they hit the registry.

- **pnpm layer** — `minimumReleaseAge: 1440` in `pnpm-workspace.yaml` blocks `pnpm install` on versions < 1 day old
- **Renovate layer** — `minimumReleaseAge: "1 day"` delays PR creation until the version has aged
- **Internal allowlist** — `@apify/*` and `@crawlee/*` (plus `apify`, `apify-client`, `crawlee`, `got-scraping`) are excluded at both layers so internal releases ship immediately

## Changes

- `package.json`: drop `"workspaces"` (moved to `pnpm-workspace.yaml`); set `packageManager` to `pnpm@10.24.0`; add `rimraf` as explicit devDep (was previously relying on npm transitive hoisting — pnpm's hoisted linker does not link undeclared transitive bins); replace `devEngines` with `npx only-allow pnpm` preinstall hook
- `lerna.json`: `npmClient: "pnpm"` so `lerna run ...` invokes pnpm
- `pnpm-workspace.yaml`: pnpm config moved here from `.npmrc`; `node-linker=hoisted` keeps a flat `node_modules` for the same debugging ergonomics as npm
- `packages/*/package.json` scripts: `npm run X` → `pnpm X`
- CI workflows: use the shared `apify/workflows/pnpm-install` composite (with pnpm-store caching). Node test matrix dropped to **[20, 22, 24]** — Node 16 is dropped because pnpm 10 needs ≥18, and Node 18 is dropped because the test runner was bumped to **vitest 4**, which declares `engines.node: ^20 || ^22 || >=24`. `engines.node` in the published packages is unchanged (no major bumps)
- `release` workflow: `pull: '--rebase --autostash'` on the bot push step to avoid non-fast-forward rejection if another commit lands while CI is running
- `renovate.json`: `minimumReleaseAge: "1 day"`, `internalChecksFilter: "strict"`, internal allowlist; drop old npm constraint
- `CONTRIBUTING.md` + `CLAUDE.md`: updated command examples

Pilot for the remaining ~13 repos in the same migration.


B4nan:
@mfori @valekjo looking at the last 100 commits, you were the most active contributors, so adding you here mostly as an FYI so you are not surprised after the merge.

I'll create oxlint/oxfmt PRs that will be targetting this branch, I want to get pnpm in first.

barjin (approved):
The changes look straightforward to me (and in line with what we have in the other repos already), so here's my ✅. 

Thanks @B4nan !

PR #623, opened 2026-04-28
chore: migrate from eslint to oxlint

B4nan:
## Summary

Replaces ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter), pulling in the shared `@apify/oxlint-config` preset and `oxlint-tsgolint` for type-aware rules. Mirrors the migration already shipped in apify-storage-local-js and apify-core.

Stacked on top of #616 (pnpm migration). Will be retargeted to `master` after #616 merges.

### What changed

- Drop `eslint`, `typescript-eslint`, `globals`, `@apify/eslint-config`
- Add `oxlint@1.62.0`, `oxlint-tsgolint@0.22.0`, `@apify/oxlint-config@^0.2.5`
- Replace `eslint.config.mjs` and `test/.eslintrc` with `oxlint.config.ts`
- `lint` / `lint:fix` now run `oxlint --type-aware`
- `lint-staged` now runs `oxlint --fix`
- `tsconfig.json` overrides `module` / `moduleResolution` to `Node16` so `oxlint-tsgolint` can parse it (build tsconfig stays on `commonjs` / `node` so tsup keeps building correctly)
- Stale `import-x/*` disable comments rewritten to oxlint's `import/*` plugin or removed where the underlying rule has no oxlint counterpart
- Auto-fix surfaced issues: switch builtin imports to `node:` protocol, drop dead Node 14 conditional in `test/log.test.ts`, fix duplicate prefix in `it('it works')`
- Allow intentional `\0` regex match in BSON key escaping (line-level disable)

### Lint output

```
$ pnpm lint
Found 0 warnings and 0 errors.
Finished in 246ms on 125 files with 115 rules using 14 threads.
```

## Test plan

- [x] `pnpm lint` — clean
- [x] `pnpm lint:fix` — clean


oxlint.config.ts
```diff
@@ -0,0 +1,44 @@
...
+            rules: {
+                'no-console': 'off',
+                'no-useless-constructor': 'off',
+                'typescript/ban-ts-comment': 'off',
+                'jest/no-conditional-expect': 'off',
+                'vitest/no-conditional-expect': 'off',
+                'jest/expect-expect': 'off',
+                'vitest/expect-expect': 'off',
+                'jest/no-standalone-expect': 'off',
+                'vitest/no-standalone-expect': 'off',
```
barjin, line 28: Some of these rules seem reasonable - is there a reason why we are disabling these? I don't see their equivalents in the `eslint` setup.

barjin (approved):
Thank you @B4nan , I only have one small nit / question ⬇️

oxlint.config.ts
B4nan: It might be all right in here, this is mostly based on `apify-core` where those triggered a lot of warnings due to false positives. I'll see if I can enable them here.

B4nan: Tested each — turned them on one at a time and counted violations:

| rule | violations | decision |
|---|---|---|
| `typescript/no-explicit-any` | 148 | keep off (opt-in case-by-case) |
| `no-param-reassign` | 40 | keep off |
| `typescript/no-empty-object-type` | 4 | enable + fix |
| `typescript/no-unsafe-declaration-merging` | 1 | enable + fix |
| `import/extensions` | 0 | enable (free) |
| `no-empty-function` | n/a | preset already has it `'off'`, drop redundant override |
| `no-use-before-define` | 2 | enable (preset config), fix |
| `no-void` | 0 | preset enables with `allowAsStatement: true`, drop our override |

Pushed as a separate commit on top of this branch so you can see the diff. The fixes are mostly removing the empty-interface declaration-merging trick (no longer needed) and switching mutually-recursive listener arrows to `function` declarations.

barjin: I meant mostly the `jest/*` and `vitest/*` rules, but it's cool we managed to shrink the main ruleset as well :+1:

camchenry: 👋 oxlint maintainer here, this might have been some noise due to splitting the Jest/Vitest rules into separate rules: https://github.com/oxc-project/oxc/issues/21664. Previously, enabling the Vitest plugin automatically enabled the Jest plugin. This is fixed in [1.63.0](https://github.com/oxc-project/oxc/releases/tag/apps_v1.63.0) so if you're only using Vitest you don't need to disable the Jest rules anymore.

PR #624, opened 2026-04-28
chore: migrate to oxfmt

B4nan:
## Summary

Replaces ad-hoc formatting with [oxfmt](https://oxc.rs/docs/guide/usage/formatter), matching the toolchain in apify-storage-local-js, apify-core, apify-web, and crawlee v4.

Stacked on top of #623 (eslint→oxlint migration). Will be retargeted to `master` after the chain merges.

### Commits

- `chore: add oxfmt tooling` — adds oxfmt 0.46.0 devDep, `.oxfmtrc.json` (4-space tabs, single quotes, 120 col), `format` / `format:check` scripts, and wires `oxfmt --write` into lint-staged
- `chore: apply oxfmt formatting` — pure `oxfmt --write` reformat across all 67 modified files; one directive move from a trailing `// eslint-disable-line` to `// eslint-disable-next-line` above a `for-in` loop (oxfmt split the trailing comment which made it apply to the wrong line)
- `ci: add format-check job` — adds `format-check` job to `test_and_release.yaml` and gates `publish` on it

## Test plan

- [x] `pnpm lint` — clean
- [x] `pnpm format:check` — clean

## Follow-up

Once this lands, add the reformat commit SHA to `.git-blame-ignore-revs`.


packages/consts/src/regexs.ts
```diff
@@ -1,6 +1,6 @@
 // Parts for building an email regex (email will be constructed as `name@domain`)
 // name parts can be alnum + some special characters
-const namePartSubRegexStr = '[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+';
+const namePartSubRegexStr = "[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+";
```
barjin, line 3: A curious question - what's the rule about single / double quotes? Here we went from single to double, but further in this file, there are multiple string literals in single quotes.

barjin (approved):
lgtm, just one question ⬇️

packages/consts/src/regexs.ts
B4nan: If you can get around the need for escaping single quotes (like on this line), double quotes are preferred.

PR #625, opened 2026-04-29
ci: use workspace protocol for internal `@apify/*` deps

B4nan:
## Summary

Fixes the publish workflow failure that surfaced after the pnpm migration (#616) merged.

### What broke

After lerna bumped versions in [run 25099449315](https://github.com/apify/apify-shared-js/actions/runs/25099449315/job/73544484277), it called `pnpm install --lockfile-only --ignore-scripts` to refresh the lockfile. pnpm then tried to fetch the newly bumped versions (e.g. `@apify/consts@^2.52.2`) from the registry — which don't exist yet because lerna hasn't published them. The publish step exited 1 with `ERR_PNPM_NO_MATCHING_VERSION`.

### Fix

Switches every internal `@apify/<package>` dependency from a registry range (e.g. `^2.52.1`) to the pnpm [`workspace:^` protocol](https://pnpm.io/workspaces#workspace-protocol-workspace). pnpm now resolves them as workspace links, and lerna replaces `workspace:^` with the actual published version at publish time — same pattern apify-storage-local-js and crawlee already use.

Lockfile diff confirms the switch is purely the resolution mechanism (`version: 2.52.1` → `version: link:../consts`); no transitive dep changes.

## Test plan

- [x] `pnpm install` — clean
- [x] `pnpm lint` — clean
- [ ] Next merge to master triggers a successful publish

### DIFF
--- a/packages/actor-memory-expression/package.json
+++ b/packages/actor-memory-expression/package.json
@@ -48,8 +48,8 @@
         "access": "public"
     },
     "dependencies": {
-        "@apify/consts": "^2.52.1",
-        "@apify/log": "^2.5.35",
+        "@apify/consts": "workspace:^",
+        "@apify/log": "workspace:^",
         "mathjs": "^15.2.0"
     }
 }
--- a/packages/input_schema/package.json
+++ b/packages/input_schema/package.json
@@ -48,9 +48,9 @@
         "access": "public"
     },
     "dependencies": {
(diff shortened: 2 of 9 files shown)


barjin (approved):
Thanks @B4nan ! I don't have much expertise in this, but it seems in line with what we do in Crawlee. 

If you're sure about it yourself, here's my obligatory ✔️

PR #626, opened 2026-04-29
chore: drop @apify/dummy-package-for-testing

B4nan:
## Summary

Removes `packages/dummy/`, the test fixture added in #442 to exercise the PR-toolkit ruleset. The package was published once at 2.2.0 in Feb 2024 and never touched since — nobody consumes it.

### Why now

Lerna tried to bump it to 2.2.1 in [run 25100324314](https://github.com/apify/apify-shared-js/actions/runs/25100324314/job/73547559292) and npm rejected with E404: the package was created before npm's trusted-publisher feature existed and has no OIDC config, while every other package publishes fine via the existing trusted-publisher setup.

Rather than wire up trusted publisher for a fixture nobody uses, just stop publishing it. The existing 2.2.0 stays on npm forever but no further versions will be pushed.

## Test plan

- [x] `pnpm install` — clean
- [x] `pnpm lint` — clean
- [x] `pnpm build` — all 14 remaining packages built


B4nan:
One last PR, this testing package was forgotten there, and it doesn't have OIDC set up, so it failed the CI.

barjin (approved):

PR #629, opened 2026-05-04
chore: drop redundant oxfmt options that .editorconfig already provides

B4nan:
## Summary

oxfmt reads `.editorconfig`, so `tabWidth` and `singleQuote` in `.oxfmtrc.json` were duplicating the existing `indent_size = 4` / `quote_type = single` from this repo's `.editorconfig`. Drop them — single source of truth, no drift.

Also dropping `trailingComma: "all"` since [it's the default](https://oxc.rs/docs/guide/usage/formatter/config-file-reference.html#trailingcomma).

`printWidth` stays in `.oxfmtrc.json` (no editorconfig equivalent we want to enforce).

Verified locally: `pnpm format:check` still clean — the diff is purely the removed lines.

### DIFF
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -1,7 +1,4 @@
 {
-    "tabWidth": 4,
-    "singleQuote": true,
-    "trailingComma": "all",
     "printWidth": 120,
     "ignorePatterns": [
         "**/*.md",


.oxfmtrc.json
```diff
@@ -1,6 +1,4 @@
 {
-    "tabWidth": 4,
-    "singleQuote": true,
     "trailingComma": "all",
     "printWidth": 120,
```
barjin, line 2: Would it make any sense to move `printWidth` to `.editorconfig.max_line_length` ([docs](https://oxc.rs/docs/guide/usage/formatter/config-file-reference.html#printwidth))?

Also, `"all"` seems to be the default for `trailingComma` ([docs](https://oxc.rs/docs/guide/usage/formatter/config-file-reference.html#trailingcomma)), maybe we can just keep the `ignorePatterns`?

barjin (approved):

.oxfmtrc.json
B4nan: Good point on trailingComma, I will keep the printWidth, since editorconfig is too dumb and it could clash with things (it had issues with long strings like URLs, and it would also not respect oxfmt-ignore comments).

PR #631, opened 2026-05-06
chore(deps): bump axios via override and ip-address in lockfile

B4nan:
## Summary
Closes the freshly-opened Dependabot alerts in `pnpm-lock.yaml`:

- **ip-address** 10.1.0 → 10.2.0 — clean transitive bump (closes #106)
- **axios** → 1.16.0 via workspace override — `nx@22.x` (latest 22.7.1 included) ships an exact `axios: 1.15.0` pin, so `pnpm update` won't bump it. Forced via `pnpm-workspace.yaml` overrides; closes #93–#105.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,6 +1,9 @@
 packages:
   - packages/*
 
+overrides:
+  axios@>=1.0.0 <2.0.0: ^1.15.2
+
 minimumReleaseAge: 1440
 minimumReleaseAgeExclude:
   - "@apify/*"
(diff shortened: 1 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -1,6 +1,9 @@
 packages:
   - packages/*
 
+overrides:
+  axios@>=1.0.0 <2.0.0: ^1.15.2
```
barjin, line 5: nit: is there a reason not to reference `^1.16.0` here?

barjin (approved):

pnpm-workspace.yaml
B4nan: no idea why claude did it this way, but 1.6 is resolved anyway

PR #637, opened 2026-05-18
feat: Add new forbidden username filters

jancurn:
Extends the forbidden usernames validation to reject usernames containing a couple of dirty keywords.

https://claude.ai/code/session_01MW39FcuEYJkGQwkGp3xBiJ

### DIFF
--- a/packages/utilities/src/utilities.ts
+++ b/packages/utilities/src/utilities.ts
@@ -670,8 +670,19 @@ const FORBIDDEN_USERNAMES_REGEXPS = [
     // Username starting with xxx-
     '(xxx-.*)',
 
-    // Username containing porn
+    // Usernames containing inappropriate/adult content keywords
     '(.*porn.*)',
+    '(.*penis.*)',
+    '(.*vagina.*)',
+    '(.*cock.*)',
+    '(.*anus.*)',
+    '(.*dick.*)',
+    '(.*pussy.*)',
+    '(.*dildo.*)',
+    '(.*nude.*)',
+    '(.*naked.*)',
+    '(.*nsfw.*)',
+    '(.*hentai.*)',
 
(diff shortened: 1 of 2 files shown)


B4nan (approved):
You might as well add more like `cunt` or at least `fuck` 😅

B4nan:
I was running this through /staff-review for the sake of testing the inline variant, and it had a few good points:

> Verification confirms the false-positive concerns are real. Summary of the worst hits:
> 
> - cock → blocks peacock, hancock, babcock, shuttlecock, cockpit, cocktail (the last two are unlikely usernames, but the surnames are not).
> - dick → blocks dickens, dickinson, dickson (extremely common surname), mobydick.
> - anus → blocks uranus, janus, manus, manuscript.
> - pussy → blocks pussycat, pussywillow.
> - nude → blocks denude, nudelman.

B4nan:
We might as well want to remove `penis` and `naked`, unfortunately 😭 

> Short answer: much smaller surface, but not zero. Two real-ish remainders:
> 
> - penis — substring of Penistone (real South Yorkshire town and surname) and openissue / openisland style compounds. The Penistone collision is the only "real surname" case left, and it's far less common than Dickens/Hancock.
> - naked — substring of snaked (past tense of snake). Unlikely username, but compounds like snakedance would also be caught.
> 
> The rest are clean:
> 
> - porn — no English-word collisions I'm aware of.
> - vagina — only matches vaginal (still NSFW). No false positives.
> - dildo — no collisions.
> - nsfw — acronym, no collisions.
> - hentai — no collisions.
> 
> So dropping cock/anus/dick/pussy/nude knocks out ~95% of the false-positive surface. The residual Penistone/snaked risk is small enough that you can either accept it or move just penis and naked to a word-boundary pattern while leaving the rest as substrings.

jancurn:
@B4nan now it should be better

PR #643, opened 2026-05-27
feat(utilities): add iterableToArray helper

MrkMrk00:
This PR adds a new utility used to collect items from an iterable (or an async iterable) into an array.

Inspiration :) https://www.php.net/manual/en/function.iterator-to-array.php

### DIFF
--- a/packages/utilities/src/streams_utilities.ts
+++ b/packages/utilities/src/streams_utilities.ts
@@ -26,3 +26,35 @@ export async function readStreamToString(stream: Readable | PassThrough, encodin
     const buffer = await concatStreamToBuffer(stream);
     return buffer.toString(encoding);
 }
+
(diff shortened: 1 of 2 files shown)


packages/utilities/src/streams_utilities.ts
```diff
@@ -26,3 +26,35 @@ export async function readStreamToString(stream: Readable | PassThrough, encodin
...
+}
+
+async function asyncIterableToArray<T>(iterable: AsyncIterable<T>): Promise<T[]> {
+    const out: T[] = [];
+    for await (const item of iterable) {
+        out.push(item);
+    }
+
+    return out;
+}
```
B4nan, line 41: this is the same as `Array.fromAsync()`, no?

MrkMrk00: :O Did not know about that one. This was my use case mainly, closing the PR. Thanks

PR #645, opened 2026-06-03
chore: added mcp origin header value

khali1:
The header was addded in mcp server: https://github.com/apify/apify-mcp-server/pull/835

Since I need to use the mcp value, to keep things clean, i need to add the value to the enum


B4nan (approved):

vdusek:
@khali1, please don't forget that we also have Python tooling. We had to discover this missing through a user bug report -  https://github.com/apify/apify-sdk-python/issues/1032.

PR #650, opened 2026-06-15
ci: add pre-publish npm drift gate

### LINKED ISSUE #649: Release pipeline: detect & prevent git-ahead-of-npm drift (post-publish verification + graceful Rekor 409)
Follow-up to #648, which manually recovered three packages that were **tagged in git but never published to npm**, silently breaking `npm install -g apify-cli` globally for ~3 days.

## What happened (root cause)

The `lerna publish` run on 2026-06-09 created git tags + GitHub releases (version phase) and then **failed mid-publish**, so `@apify/utilities@2.33.0`, `@apify/input_secrets@1.2.44`, and `@apify/markdown@3.0.68` never reached npm. Nothing detected the gap. On 2026-06-12 a dependent (`@apify/input_schema@3.28.15`) was published with `workspace:^` resolving to the in-repo (unpublished) `input_secrets@1.2.44`, exposing it as `latest` → every fresh install hit `ETARGET`.

The specific trigger was an npm/sigstore provenance error:
```
@apify/utilities: TLOG_CREATE_ENTRY_ERROR — (409) an equivalent entry already exists in the transparency log
```
The Rekor entry was integrated at 11:28:37Z, ~70s into the run — i.e. `make-fetch-happen`'s retry inside `@sigstore/sign`'s `TLogClient.createEntry` **re-POSTed an attestation that had already succeeded** (lost-ack on a non-idempotent create), and Rekor refused the duplicate. lerna then aborted the whole multi-package publish (worsened by its `errno "undefined" is not a valid exit code` bug), leaving tags ahead of npm.

## Why this keeps being dangerous

- A 409 is just **one** way a publish can half-fail — a 5xx, rate-limit, auth blip, or aborted run all leave the same silent git-ahead-of-npm drift.
- Because tags exist, **re-running the pipeline no-ops** (`lerna changed` = 0), and `lerna publish from-package` re-attempts the *same* versions whose digests are permanently poisoned by the immutable Rekor 409. Only a fresh version escapes — which is the manual workaround #648 had to do.

### 1. Post-publish verification (highest priority)
After `lerna publish` in `publish_to_npm.yaml`, assert every package's current version actually resolves on npm, and fail the job otherwise:
```bash
pnpm exec lerna ls --json \
  | jq -r '.[] | "\(.name)@\(.version)"' \
  | xargs -I{} sh -c 'npm view {} version >/dev/null 2>&1 || { echo "MISSING ON NPM: {}"; exit 1; }'
```
Turns *any* drift into an immediate red build instead of a latent global breakage.

### 2. Handle the Rekor 409 gracefully (root cause of the trigger)
Bump npm in the publish job so the bundled `sigstore`/`@sigstore/sign` treats a duplicate-entry 409 as success (newer versions already do this — the UUID is content-addressed, so the existing entry is provably ours). Do **not** swallow it at the YAML layer (`|| true`/grep) — that would mask real failures too.

### 3. (Optional) Reconcile at the start of each release
Run `lerna publish from-package` (or the verification in #1) at the **beginning** of a release so any pre-existing drift is healed before new dependents bake `workspace:^` into their published manifests.

## Acceptance
- A release that bumps+tags but fails to publish any package **fails the workflow** (verified by the check in `#1`).
- A lost-ack Rekor retry no longer aborts the publish (verified by the npm bump in `#2`).

Refs: #648

l2ysho:
Adds a pre-publish step to publish_to_npm.yaml that fails the release fast if a publishable package's tagged version s missing from npm — drift left behind when a prior run bumped and tagged but never published. Packages this run is about to release are skipped, since they legitimately aren't on npm yet (new versions or recovery re-publishes).

closes #649


DaveHanns (approved):
LGTM 👍

vladfrangu (approved):

PR #654, opened 2026-06-19
fix: fix incorrect types

mvolfik:
wtf (look at how the params are checked a few lines below)

(introduced in #137, 2021)


B4nan (approved):
not sure what the formatter dislikes, otherwise looks good

PR #661, opened 2026-07-03
feat: memory expression allow comparing text

### LINKED ISSUE #660: actor-memory-expression: Enable matching on string values
String enums are pretty common way to add extra features that can require more CPU, e.g. `"dataOutput": { "type": "string", "enum": ["basic", "full detail"] }`. Currently, there is no way to match on string values. An ugly workaround is to match on string value length, e.g. `get(input, "dataOutput.length") > 5` but that won't work if there are many values of similar length.

daniil-poletaev:
This PR extends MathJS operations to support `compareText()`.

Closes #660

### DIFF
--- a/packages/actor-memory-expression/src/memory_calculator.ts
+++ b/packages/actor-memory-expression/src/memory_calculator.ts
@@ -2,9 +2,11 @@
 import {
     addDependencies,
     andDependencies,
+    compareTextDependencies,
     compileDependencies,
     create,
     divideDependencies,
+    equalDependencies,
     evaluateDependencies,
     maxDependencies,
     minDependencies,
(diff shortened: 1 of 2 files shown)


nmanerikar (approved):
LGTM

nmanerikar:
Were there any security concerns about `equalDependencies` and `unequalDependencies` previously? These will now be generally available (not just in conjunction with `compareText`)

daniil-poletaev:
> Were there any security concerns about equalDependencies and unequalDependencies previously? These will now be generally available (not just in conjunction with compareText)

I don't remember any security concers regarding equalDependencies or unequalDependencies. We just wanted to limit amount of operations we support to be safer and add them one by one

PR #669, opened 2026-07-15
feat(timeout): add `extendTimeout` to push back a running deadline

B4nan:
Adds `extendTimeout(extraMillis)`, for cases where the time a handler needs is only known once it is already running — e.g. a listing page that turns out to have a lot of pages to scroll through. Needed for [apify/crawlee#1485](https://github.com/apify/crawlee/issues/1485) and [apify/crawlee#2951](https://github.com/apify/crawlee/issues/2951), where a request handler needs to be able to ask for more time from the inside.

```ts
await addTimeoutToPromise(async () => {
    const pageCount = await countPages();
    tryCancel();
    extendTimeout(pageCount * 10_000);
    await scrapeAllPages();
}, 30_000, 'Handler timed out!');
```

### How

Nested calls used to share a single store object, so there was nowhere to hang a per-layer handle on — the timer was only ever reachable from the closure of the `addTimeoutToPromise` call that created it. Each call now gets a store of its own, carrying a handle to its own timer.

Two behaviours worth calling out:

- **Enclosing timeouts are extended by the same amount by default.** Otherwise a nested handler asking for more time would just be killed moments later by an outer timeout it cannot see (in crawlee's case, an internal safety net the handler author never set and cannot observe). Pass `propagate: false` to extend only the innermost one, keeping an enclosing timeout as a hard limit on the total time.
- **Extending an already timed out or finished handler is a no-op.** Cancellation is cooperative, so a handler that ignores `tryCancel()` keeps running after it timed out — without the guard it could arm a timer for an already settled promise, achieving nothing but holding the event loop open.

### Compatibility

No public API changed: `addTimeoutToPromise`, `tryCancel`, `TimeoutError` and `AbortContext` are all untouched, and `TimeoutFrame` stays internal. The `AbortController` is still inherited from the enclosing frame, so `tryCancel()` and the existing nesting semantics behave exactly as before.

The only observable difference is that `storage.getStore()` returns a different object per nesting level rather than one shared object. Since `storage` is exported, each frame prototypally inherits from the enclosing one, so reads of anything stashed on an outer context still resolve.

Verified against crawlee (aliasing the package to this branch): 676 core tests and 94 browser-pool tests pass. That run also caught a regression this branch originally had — passing a *named* function to `setTimeout` replaced the `Timeout._onTimeout` frame that crawlee matches on when logging timeout errors. Fixed, and covered by a new test here.

### DIFF
--- a/packages/timeout/src/index.ts
+++ b/packages/timeout/src/index.ts
@@ -5,6 +5,18 @@ export interface AbortContext {
 }
 
 /**
+ * A single `addTimeoutToPromise` call. Nested calls share the `cancelTask` controller of the
+ * enclosing frame (so `tryCancel()` behaves the same at any depth), but each call gets its own
+ * frame, which is what lets {@apilink extendTimeout} tell the layers apart.
+ */
+interface TimeoutFrame extends AbortContext {
+    /** The enclosing frame, when this call is nested inside another `addTimeoutToPromise`. */
+    parent?: TimeoutFrame;
+    /** Pushes this frame's deadline back. Becomes a no-op once the frame times out or settles. */
+    extend?: (extraMillis: number) => void;
+}
+
+/**
  * `AsyncLocalStorage` instance that is used for baring the AbortContext inside user provided handler.
  * We can use it to access the `AbortContext` instance via `storage.getStore()`, and there we can access
  * `cancelTask` instance of `AbortController`.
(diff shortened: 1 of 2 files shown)


packages/timeout/src/index.ts
```diff
@@ -4,6 +4,18 @@ export interface AbortContext {
...
 
+/**
+ * A single `addTimeoutToPromise` call. Nested calls share the `cancelTask` controller of the
+ * enclosing frame (so `tryCancel()` behaves the same at any depth), but each call gets its own
+ * frame, which is what lets {@link extendTimeout} tell the layers apart.
+ */
+interface TimeoutFrame extends AbortContext {
+    /** The enclosing frame, when this call is nested inside another `addTimeoutToPromise`. */
+    parent?: TimeoutFrame;
+    /** Pushes this frame's deadline back. Returns `false` once the frame timed out or settled. */
```
barjin:
```suggestion
    /** Pushes this frame's deadline ahead. Returns `false` if the frame has already timed out or settled. */
```

barjin (approved):
One docs nit, otherwise lgtm. Thanks @B4nan !

PR #684, opened 2026-08-26
feat: publish ESM-only packages and require Node.js 22+

B4nan:
All packages now ship a single ESM build in `dist/` instead of the dual CJS/ESM layout, and declare `engines.node >=22`, matching the upcoming crawlee v4, apify SDK v4 and apify-client v3. CommonJS consumers on Node 22+ can still `require()` the packages through native `require(esm)`.

Details worth reviewing:

- `tsup` produces one `esm` build per package; the shared config lost its CJS and IIFE branches
- `tsconfig` moved to `module: nodenext`, so relative imports in sources carry `.js` extensions and JSON imports use `with { type: 'json' }`
- ajv imports switched to the named class exports (`Ajv`, `Ajv2019`), since default imports of CJS modules no longer type-check under `nodenext`
- the `require('node:crypto')` fallback in the `@apify/utilities` hmac helper is gone, Node 22 always has `globalThis.crypto.subtle`
- the `ow` interop banner hack in `@apify/input_secrets` stays until we bump `ow` to an ESM version
- the `json_schemas` build scripts run with plain `node` instead of `tsx` (dev dep dropped). Node's type stripping needs literal `.ts` import specifiers, so the unpublished `scripts/` and `tools/` files use `.ts` extensions; package sources keep the usual `.js` convention

Stacked on #683.

BREAKING CHANGE: all packages are ESM-only and require Node.js 22+.


packages/input_schema/src/utilities.ts
```diff
@@ -1,13 +1,13 @@
 import { parse } from 'acorn-loose';
 import type { ValidateFunction } from 'ajv';
-import type Ajv from 'ajv/dist/2019';
+import type { Ajv2019 as Ajv } from 'ajv/dist/2019.js';
```
vladfrangu, line 3: The fact this works shocks me.

vladfrangu (approved):

PR #685, opened 2026-08-26
feat(input_secrets): replace ow with zod validation

B4nan:
Also extracts the wrapper into a new `@apify/validations` package (review suggestion), which `input_secrets` consumes and re-exports, so crawlee and apify-client can drop their vendored copies once this ships.


packages/validations/src/validation.ts
```diff
@@ -0,0 +1,161 @@
+import { z } from 'zod';
+
+/** Formats a zod issue path like `groups[0]` or `countryCode`. */
+function formatIssuePath(path: readonly PropertyKey[]): string {
```
vladfrangu, line 4: Might be time to make the apify/validations pr in this batch of PRs so we unite this once and for all :D

B4nan: Done, extracted into a new `@apify/validations` package in this PR. `input_secrets` consumes and re-exports it, so crawlee and the client can drop their vendored copies once this ships.

PR #686, opened 2026-08-26
ci: support publishing beta prereleases from non-default branches

B4nan:
Every push to `next` now automatically publishes the changed packages as beta prereleases under the `next` npm dist-tag (`npm install @apify/<pkg>@next`), so the upcoming majors can be tested end to end (crawlee e2e suite and elsewhere) before the fast-forward to `master`. The `next` branch stays a long-lived prerelease channel afterwards, usable for betas of patch/minor bumps the same way.

Like in crawlee, betas leave no trace in git: no version commits, no tags, no GitHub releases, no changelog updates. The mechanics:

- `lerna version --conventional-prerelease --preid beta --no-push --no-git-tag-version --no-changelog` computes the bump level from conventional commits since the last stable release (accumulated breaking changes produce one major, e.g. `3.0.0-beta.N`, and the four 0.x packages go to `1.0.0-beta.N`)
- since git records nothing, lerna alone would always produce `beta.0`; `scripts/resolve-beta-versions.ts` rewrites the prerelease number to the next free one according to the npm registry (crawlee's approach, adapted for independent versioning)
- a throwaway local commit keeps the tree clean for `lerna publish from-package --dist-tag next`; it is never pushed
- stable releases from `master` are untouched: the same conventional commits then produce the stable versions, so no graduation step is needed

Verified on a scratch clone: the lerna invocation skips tags/commits/releases as expected and proposes the right versions for all 14 packages, and the beta-number logic was checked against real registry data. The manual `workflow_dispatch` path also accepts the `prerelease` flag for betas from other branches. Documented in CONTRIBUTING.md. Stacked on the zod PR.

### DIFF
--- a/.github/workflows/publish_to_npm.yaml
+++ b/.github/workflows/publish_to_npm.yaml
@@ -17,6 +17,10 @@ permissions:
     id-token: write # Required for OIDC
     contents: write # Required to push changelog/lock file updates
 
+# Serialize publishes: concurrent runs would compute the same beta prerelease numbers from the registry
+concurrency:
+    group: publish_to_npm
+
 jobs:
     publish_to_npm:
         name: Publish to NPM
--- a/.github/workflows/publish_to_npm.yaml
+++ b/.github/workflows/publish_to_npm.yaml
@@ -7,6 +7,11 @@ on:
                 description: Git ref to publish (branch, tag, or commit SHA)
                 required: true
                 type: string
+            prerelease:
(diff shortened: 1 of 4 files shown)


vladfrangu (approved):
Petition to make a PR dropping lerna, or an epic for me to tackle that across repos :D

B4nan:
Agreed, but epic material rather than this stack: the whole release flow here (versioning, changelogs, topological publish) leans on lerna.

PR #690, opened 2026-08-26
chore: update dev dependencies

B4nan:
Dev dependency bumps: lerna 10 (needs Node >=22.13, satisfied by the new floor), lint-staged 17, commitlint 21, strip-ansi 7, oxlint 1.80 + oxfmt 0.65 + `@apify/oxlint-config` 0.3, and oxlint-tsgolint 7 (verified type-aware rules still fire, with TypeScript staying on 5.9). Also drops the deprecated `@types/git-url-parse` (v16 ships its own types) and pins `@types/node` to 22 to match the engines floor.

TypeScript 7 itself does not work yet: tsup's declaration bundling (rollup-plugin-dts) needs the JS compiler API that tsgo does not expose, so the compiler stays on 5.9 for now. One new oxlint rule (`import/namespace`) is disabled for tests, where data-driven suites access imported namespaces with computed keys. Stacked on the runtime-independence PR.


package.json
```diff
@@ -49,25 +49,24 @@
...
-        "lint-staged": "^16.0.0",
+        "lerna": "^10.0.0",
+        "lint-staged": "^17.0.0",
         "nock": "^14.0.0",
-        "oxfmt": "0.46.0",
-        "oxlint": "1.62.0",
-        "oxlint-tsgolint": "0.22.0",
+        "oxfmt": "0.65.0",
+        "oxlint": "1.80.0",
+        "oxlint-tsgolint": "7.0.2001",
```
vladfrangu, line 67: Why are these deps fixed and not ranged?

B4nan: Deliberate, and they were pinned before this PR too: oxlint/oxfmt minors can change lint or format output, which breaks format:check in CI until everyone updates in lockstep, and oxlint-tsgolint has to match the oxlint version exactly. Renovate still bumps the pins.

PR #691, opened 2026-08-26
chore: replace tsup with tsc and adopt TypeScript 7

B4nan:
Drops tsup and compiles with plain `tsc` from TypeScript 7 (the native Go compiler), the same setup crawlee v4 uses. Bundling bought us nothing anymore: dependencies were always external, the exports maps from #684 already forbid deep imports, and tsc copies the imported JSON schemas into `dist/` by itself. tsup was also the one thing blocking TS 7, since its declaration bundling needs the TypeScript JS compiler API that tsgo does not provide (tracked in egoist/tsup#1408).

What changed:

- per-package `compile` is `tsc -p tsconfig.build.json`; all `tsup.config.ts` files and the tsup dev dependency are gone
- dist is multi-file now (per-module `.js` + `.d.ts` + sourcemaps) instead of one bundled file; consumption through the exports maps is unchanged, verified by importing every built package under plain Node and a `require(esm)` check
- `json_schemas` emits `dist/src` + `dist/schemas` (its JSON lives outside `src/`), so its entry moved to `./dist/src/index.js`
- TS 7 needed two tsconfig adjustments: explicit `rootDir` per package, and an explicit `"types": ["node"]` (7.x no longer picks `@types/node` up automatically); `skipLibCheck` is now on, which every tsc-based sibling repo also uses
- if we ever want bundling back, it can return without changing how the packages are consumed

The commit is a `chore` on purpose: the dist layout change ships together with the pending major bumps. Stacked on the dev-deps PR.


tsconfig.build.json
```diff
@@ -4,8 +4,11 @@
     "target": "es2023",
     "moduleResolution": "nodenext",
     "declaration": true,
-    "sourceMap": false,
+    "sourceMap": true,
+    "inlineSources": true,
```
vladfrangu, line 8: I don't think we should inline sourcemaps

packages/timeout/package.json
vladfrangu, line 10: Why did only one package change the paths to `./dist/src/*`?

vladfrangu (changes requested):

tsconfig.build.json
B4nan: These are not inline source maps, the maps stay separate .map files. inlineSources only embeds the TS sources into the map's sourcesContent. We publish just dist/, so the ../src/*.ts paths in the maps point outside the tarball and stack traces would have nothing to show without it. tsup embedded the sources the same way.

packages/timeout/package.json
B4nan: json_schemas imports ../schemas/*.json from outside src/, so its rootDir is the package root and tsc mirrors the layout as dist/src + dist/schemas. Every other package has all inputs under src/, so they emit flat dist/*.

PR #692, opened 2026-08-26
feat(payment_qr_codes): generate QR codes with etiket instead of qrcode

B4nan:
Replaces `qrcode` with `etiket` in `@apify/payment_qr_codes`. `qrcode` shipped `pngjs`, `dijkstrajs` and the whole `yargs` CLI parser to every consumer and was pinned + renovate-ignored; `etiket` has zero dependencies and no Node builtins, which makes this the second fully runtime-independent package. The renovate ignore entry is gone.

Not a breaking change: the function keeps returning a PNG data URL, and with `ecLevel`/`moduleSize`/`margin` matching the old defaults the output is pixel-identical to what `qrcode` produced (verified by decoding the generated code back to the SPAYD payload). The branch started with `uqr` and SVG output; review feedback pointed at etiket, which keeps the PNG format. Stacked on the tsc/TS7 PR.


packages/payment_qr_codes/src/czk_payment_qr_codes.ts
```diff
@@ -1,4 +1,4 @@
-import qrcode from 'qrcode';
+import { renderSVG } from 'uqr';
```
vladfrangu: uqr also recommends https://github.com/productdevbook/etiket which seems to support native 0-dep png output which would make this less breaking, thoughts?

B4nan: Good call, switched to etiket. PNG output is back and pixel-identical to what the qrcode package produced (verified by decoding the generated code and comparing symbol geometry), so this is no longer a breaking change at all.

PR #693, opened 2026-08-26
feat(git): replace git-url-parse with a small built-in parser

B4nan:
`@apify/git` only needs the host and the `owner/repo` pair from the repo URL, so a `URL` constructor plus one regex for the scp-like form (`git@host:owner/repo.git`) replaces `git-url-parse`. That drops the `git-url-parse` -> `git-up` -> `parse-url` -> `parse-path` -> `protocols` chain, five packages with a history of ReDoS advisories (the broken `normalize-url` typings we kept hitting in builds came from there too). The package now has zero dependencies.

Existing scp-style tests pass unchanged; new tests cover https, ssh and Bitbucket web URL forms (including the path suffix after `owner/repo` that the old code special-cased). Stacked on the uqr PR.

### DIFF
--- a/test/git.test.ts
+++ b/test/git.test.ts
@@ -260,45 +260,33 @@ describe('convertRelativeImagePathsToAbsoluteInReadme()', () => {
         ).toEqual(testMarkdown);
     });
 
-    it('works with https repo URLs', () => {
-        const testMarkdown = '![img](./img.jpg)';
-
-        expect(
-            convertRelativeImagePathsToAbsoluteInReadme({
-                readme: testMarkdown,
-                gitRepoUrl: 'https://github.com/apify/test-repo.git',
-            }),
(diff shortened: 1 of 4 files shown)


test/git.test.ts
```diff
@@ -259,4 +259,46 @@ describe('convertRelativeImagePathsToAbsoluteInReadme()', () => {
...
+                gitBranchName: 'main',
+            }),
+        ).toEqual('![img](https://bytebucket.org/apify/test-repo/raw/main/img.jpg)');
+
+        expect(
+            convertRelativeImagePathsToAbsoluteInReadme({
+                readme: testMarkdown,
+                gitRepoUrl: 'ssh://git@gitlab.com/apify/test-repo.git',
+            }),
+        ).toEqual('![img](https://gitlab.com/apify/test-repo/-/raw/master/img.jpg)');
```
vladfrangu: Just small nit, this could've been an `expect.each()`

vladfrangu (approved):

PR #699, opened 2026-09-03
feat: the v3 major releases (ESM-only, Node 22+)

B4nan:
Lands the v3 majors from the `next` integration branch: ESM-only packages, Node 22+, TypeScript 7 builds, zod validation, the new `@apify/validations` package and the dependency cleanups. Thirteen linear commits, to be merged with rebase and merge so the conventional messages and BREAKING CHANGE footers drive the version bumps. Betas from this branch were verified in crawlee, apify-sdk-js and apify-core.


.github/workflows/test_and_release.yaml
```diff
@@ -2,9 +2,9 @@ name: Check & Release
 
 on:
     push:
-        branches: [ master, renovate/** ]
+        branches: [ master, next, renovate/** ]
     pull_request:
-        branches: [ master ]
+        branches: [ master, next ]
```
barjin, line 7: Are we planning to keep the `next` branch? If so, shall we add a GitHub rule protecting this branch (as push to `next` will publish an npm package now)?

Image

barjin (approved):
Thanks @B4nan ! 

I suppose this is a stack of the separate PRs from before, right? The changes look alright to me, just one nit about branch naming / protection ⬇️

.github/workflows/test_and_release.yaml
B4nan: Hmm, good idea, we will keep it for future so people can ship dev versions first if they want.

B4nan:
Landing via fast-forward push instead (org ruleset bypass); content is identical to the reviewed head.

